PR #1, opened 2023-04-20
DEVEXP-386 Can now use connector in PySpark

rjrudin:
Had to move some things around:

- Moved MarkLogicTableProvider to `com.marklogic.spark` and renamed it `DefaultSource` to follow the Spark 2 convention, which PySpark seems to rely on.
- Renamed `MarkLogicReader` to `MarkLogicTable` as we'll be adding "write" support to that soon.
- Added details to the CONTRIB file on how to use PySpark.


BillFarber (approved):
LGTM

anu3990 (approved):

PR #4, opened 2023-04-25
DEVEXP-378 Now using multiple hosts in a cluster

rjrudin:
I wrote up DEVEXP-406 to address what happens when a particular host stops working; DMSDK has support for that - i.e. that client is removed and not used again. 

To fully test this, we need a multi-node cluster, and ideally with at least 1 node not having a forest for the database. That will be addressed in a follow-up PR that adds a docker-compose file for building such a cluster and doing manual testing with it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,11 +8,18 @@ plugins {
 group 'org.example'
 version '1.0-SNAPSHOT'
 
+java {
+  sourceCompatibility = 1.8
+  targetCompatibility = 1.8
+}
+
 repositories {
   // Temporarily here for resolving Java Client 6.2-SNAPSHOT.
   maven {
     url "https://nexus.marklogic.com/repository/maven-snapshots/"
   }
+  mavenLocal()
+
   mavenCentral()
 }
 
@@ -22,7 +29,7 @@ dependencies {
 
   // Makes it possible to use lambdas in Java 8 to implement Spark's Function1 and Function2 interfaces
   // See https://github.com/scala/scala-java8-compat for more information
-  implementation ("org.scala-lang.modules:scala-java8-compat_2.12:1.0.2") {
+  implementation("org.scala-lang.modules:scala-java8-compat_2.12:1.0.2") {
     // Prefer the Scala libraries used within the user's Spark runtime.
     exclude module: "scala-library"
   }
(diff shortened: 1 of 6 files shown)


src/main/java/com/marklogic/spark/reader/HostUtil.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Determines hosts for distributing load. A "data host" (or "d-node") is one with at least forest associated with the
+ * database that the given Java Client points to.
+ * <p>
+ * Mostly copied from BatcherImpl and RowBatcherImpl in the Java Client.
```
rjrudin, line 18: This is the copy/paste job from the Java Client. It would be nice to make that code reusable, particularly the logic for removing a client when it stops responding. That would be a significant refactor though because the logic is buried a bit within DMSDK.

src/main/java/com/marklogic/spark/reader/MarkLogicPartitionReader.java
```diff
@@ -142,4 +149,14 @@ private JacksonParser newJacksonParser(StructType schema) {
         final Seq<Filter> filters = JavaConverters.asScalaIterator(new ArrayList().iterator()).toSeq();
         return new JacksonParser(schema, jsonOptions, allowArraysAsStructs, filters);
     }
+
+    private RowManager selectRowManagerFromDataHosts() {
```
rjrudin, line 153: This is the logic for choosing a host to use. I've verified it works against a multi-node cluster, and I'm going to provide test support for that in a follow-up PR that puts a docker-compose file in place for building a 3-node cluster.

build.gradle
```diff
@@ -32,6 +39,7 @@ dependencies {
   testImplementation "ch.qos.logback:logback-classic:1.3.5"
   testImplementation "org.slf4j:jcl-over-slf4j:1.7.36"
   testImplementation "org.skyscreamer:jsonassert:1.5.1"
+  testImplementation "com.marklogic:ml-app-deployer:4.5.1"
```
rjrudin, line 42: Added this for easy calls to the Manage API for e.g. getting a list of host names.

BillFarber (dismissed):
LGTM

BillFarber (dismissed):

BillFarber (approved):

PR #6, opened 2023-04-26
Added 3-node testing support via Docker

rjrudin:
See the CONTRIB file on instructions on how to try this out locally.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,33 @@ You can then run all the tests:
 
     ./gradlew test
 
+# Running the tests against a 3-node cluster
+
+This project includes a `docker-compose.yaml` file for standing up a 3-node MarkLogic cluster. This is useful for 
+manually testing the connector's support for distributing requests to multiple hosts in a cluster. The automated 
+tests can also be run against this cluster too.
+
+To run the automated tests, first perform the following steps to set up a 3-node cluster and a Docker image 
+containing all of this project's Gradle dependencies:
+
+1. [Install Docker](https://docs.docker.com/get-docker/).
+2. Ensure that you don't have a MarkLogic instance running locally.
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -15,6 +15,33 @@ You can then run all the tests:
...
+manually testing the connector's support for distributing requests to multiple hosts in a cluster. The automated 
+tests can also be run against this cluster too.
+
+To run the automated tests, first perform the following steps to set up a 3-node cluster and a Docker image 
+containing all of this project's Gradle dependencies:
+
+1. [Install Docker](https://docs.docker.com/get-docker/).
+2. Ensure that you don't have a MarkLogic instance running locally.
+3. Run `./gradlew dockerUp` (Gradle tasks are included as shortcuts for running Docker commands). This will start up 
+   the 3-node cluster, with one host listening on the standard ports from 8000 to 8002.
```
BillFarber, line 30: Is it worth reminding people that this means a local MarkLogic would need to be shutdown?
If 8000-8002 are currently in use, Docker will fail.

BillFarber (approved):

CONTRIBUTING.md
rjrudin: Hopefully they read step 2. ;)

docker-compose.yaml
```diff
@@ -0,0 +1,74 @@
+version: '3.6'
+name: marklogic_spark
+
+services:
+  bootstrap_3n:
+    image: "marklogicdb/marklogic-db:11.0.2-centos-1.0.2"
+    container_name: bootstrap_3n
```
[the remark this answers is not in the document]
rjrudin: I think the container names can be changed at any point - the containers/volumes can be deleted at any point and then rebuilt. I just stuck with these names as they're used in the MarkLogic Docker docs for the 3 nodes.

PR #10, opened 2023-04-27
DEVEXP - 373 : Run a Spark 3 program that writes data to MarkLogic

anu3990:


src/main/java/com/marklogic/spark/writer/MarkLogicDataWriter.java
```diff
@@ -0,0 +1,67 @@
...
+    Map<String, String> map;
+    public MarkLogicDataWriter(Map<String, String> map) {
+        this.map = map;
+    }
+
+    @Override
+    public void write(InternalRow record) throws IOException {
+        System.out.println("************ Reached MarkLogicDataWriter.write ****************");
+      // TODO: Find a way to write based on the schema passed in by the user
+        DatabaseClient db = DatabaseClientFactory.newClient(map.get("spark.marklogic.client.host"), Integer.valueOf(map.get("spark.marklogic.client.port")),
```
rjrudin, line 41: Use the line of code that's in `ReadContext` since we already have a good way of creating this:

    DatabaseClient client = DatabaseClientFactory.newClient(propertyName -> properties.get("spark." + propertyName));

I imagine we'll make a `WriteContext` to carry about state related to writing documents, but no need for that yet.

src/main/java/com/marklogic/spark/writer/MarkLogicDataWriter.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+
+public class MarkLogicDataWriter implements DataWriter<InternalRow> {
+    Map<String, String> map;
+    public MarkLogicDataWriter(Map<String, String> map) {
+        this.map = map;
+    }
+
+    @Override
+    public void write(InternalRow record) throws IOException {
```
rjrudin, line 38: There's a `convertRowToJSONString` method in `HubDataWriter` in the DHF Spark connector - let's use that for now to convert this into a JSON document. We'll probably change what it's doing, but it seems like a good starting point for converting the row into JSON.

src/main/java/com/marklogic/spark/writer/MarkLogicWriteTableProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.spark.sql.connector.catalog.TableCapability;
+import org.apache.spark.sql.connector.write.LogicalWriteInfo;
+import org.apache.spark.sql.connector.write.WriteBuilder;
+import org.apache.spark.sql.types.StructType;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class MarkLogicWriteTableProvider implements SupportsWrite {
```
rjrudin, line 28: Instead of creating this class, make the existing `MarkLogicTable` class implement `SupportsWrite`.

src/test/java/com/marklogic/spark/MarkLogicSparkWriteDriver.java
```diff
@@ -0,0 +1,76 @@
...
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.DatabaseClientFactory;
+import com.marklogic.client.document.TextDocumentManager;
+import com.marklogic.spark.writer.MarkLogicWriteDataSource;
+import org.apache.spark.sql.*;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+
+public class MarkLogicSparkWriteDriver {
```
rjrudin, line 27: We already have test plumbing to reuse, so let's make use of that - make this into a `WriteRowsTest` that extends `AbstractIntegrationTest`. Can use `newSparkSession()` to create a Spark session and use that to read from the CSV. And then write the results to MarkLogic in a hardcoded collection, and then verify that the collection has N documents in it.

To make the test repeatable, modify `deleteDocumentsBeforeTestRuns` in `AbstractIntegrationTest` to delete the hardcoded collection that this test uses - e.g. 

```
public void deleteDocumentsBeforeTestRuns() {
    getDatabaseClient().newServerEval().javascript("xdmp.collectionDelete('my-test-data')").evalAs(String.class);
}
```

src/main/java/com/marklogic/spark/writer/MarkLogicWriteDataSource.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.spark.sql.connector.catalog.Table;
+import org.apache.spark.sql.connector.catalog.TableProvider;
+import org.apache.spark.sql.connector.expressions.Transform;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+
+import java.util.Map;
+public class MarkLogicWriteDataSource implements TableProvider {
```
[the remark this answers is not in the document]
anu3990: Tried to use DefaultSource but since they have the same overridden methods, there are conflicts. We can create a story to refactor it later.

src/main/java/com/marklogic/spark/writer/MarkLogicDataWriter.java
anu3990: Currently the "record" is not extracted. We can handle this later when we fine tune for different types of documents.

src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
```diff
@@ -0,0 +1,73 @@
...
+import com.marklogic.client.DatabaseClient;
+import com.marklogic.client.DatabaseClientFactory;
+import com.marklogic.client.document.TextDocumentManager;
+import com.marklogic.spark.AbstractIntegrationTest;
+import org.apache.spark.sql.*;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+
+public class WriteRowsTest extends AbstractIntegrationTest {
```
rjrudin, line 27: This failed for me when I ran it because it's depending on port 8012. Our test app is running on port 8016 though. 

But it's easy to convert this to a real test and get it working - just use the following, which turns `main` into a test method and then uses the config from the parent class to connect to MarkLogic:

```
@Test
    void test() {
        SparkSession sparkSession = newSparkSession();
        Dataset<Row> dataset = sparkSession.read().option("header", true)
            .format("csv")
            .csv("src/test/ml-data/data.csv"); // Path from repository root

        // TODO: The dataset passed in should be available to read (TableProvider or DataWriter?)
        System.out.println(dataset); // [docNum: string, docName: string]
        StringBuffer headers = new StringBuffer();
        for (int i = 0; i < dataset.schema().fields().length; i++) {
            headers.append(dataset.schema().fields()[i].name());
            headers.append(",");
        }
        System.out.println("************ Starting MarkLogicSparkWriteDriver ****************");
        DataFrameWriter<Row> dataFrameWriter = dataset
            .write()
            .format(MarkLogicWriteDataSource.class.getName())
            .option("spark.marklogic.client.host", testConfig.getHost())
            .option("spark.marklogic.client.port", testConfig.getRestPort())
            .option("spark.marklogic.client.username", testConfig.getUsername())
            .option("spark.marklogic.client.password", testConfig.getPassword())
            .option("spark.marklogic.client.authType", "digest")
            // TODO: The below schema passed in should be readable by the inferschema method.
            .option("schema", headers.toString())
            .mode(SaveMode.Append); // mode is needed, else the doc is not written

        dataFrameWriter.save();
        verifyDocIsWritten();
    }

    private void verifyDocIsWritten() {
        TextDocumentManager textDocumentManager = getDatabaseClient().newTextDocumentManager();
        assertNotNull(textDocumentManager.read("doc1.txt"));
    }
```

src/main/java/com/marklogic/spark/writer/MarkLogicWriteDataSource.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public StructType inferSchema(CaseInsensitiveStringMap options) {
+       // TODO: Need a story to read the schema passed in since this method is needed and it cannot return null.
```
rjrudin, line 30: This is interesting - we of course don't require any schema in order for data to be written. And in our Spark 2 connector, it looks like  `StructType schema` was always passed in via `DefaultSource`. Ideally, we can support a way for the user not to have to provide a schema, but maybe that won't be possible.

rjrudin (changes requested):
Looks good, just convert the test to be a real test so that it uses the test config and passes.

src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
anu3990: Updated the test, but it throes "NullPointerException" on testConfig.getHost(). Maybe "testConfig" is not populated.

rjrudin (approved):

PR #11, opened 2023-04-29
Trying out load balancer with aggregated logging

rjrudin:


rjrudin:
Just a prototype for now.

PR #23, opened 2023-05-09
Dirty prototype for streaming

rjrudin:


src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
```diff
@@ -34,6 +38,36 @@ public class WriteRowsTest extends AbstractIntegrationTest {
...
+            .option(Options.WRITE_BATCH_SIZE, 10)
+            .option(Options.WRITE_THREAD_COUNT, 1)
+            .option("spark.marklogic.client.host", testConfig.getHost())
+            .option("spark.marklogic.client.port", testConfig.getRestPort())
+            .option("spark.marklogic.client.username", "spark-test-user")
+            .option("spark.marklogic.client.password", "spark")
+            .option("spark.marklogic.client.authType", "digest")
+
+            // This should really be a new temp directory
+            .option("checkpointLocation", "/Users/rrudin/spark-temp/")
```
rjrudin, line 62: This is a dirty hack - it's up to the user to provide this (our connector doesn't care about it). For a JUnit test, we can just use a random temp directory - I think JUnit 5 has support for just such a thing.

src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
```diff
@@ -34,6 +38,36 @@ public class WriteRowsTest extends AbstractIntegrationTest {
...
 
+    @Test
+    void streaming() throws Exception {
+        StreamingQuery stream = newSparkSession().readStream()
+            .schema(new StructType()
+                .add("docNum", DataTypes.IntegerType)
+                .add("docName", DataTypes.StringType))
+            .option("header", true)
+            .format("csv")
+            // Spark complains if this is a single file path - ??
```
rjrudin, line 49: I didn't dig into why this is - but Spark complained if I did "data.csv", but is fine with "data*.csv".

src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
```diff
@@ -34,6 +38,36 @@ public class WriteRowsTest extends AbstractIntegrationTest {
 
     private final static String EXPECTED_COLLECTION = "my-test-data";
 
+    @Test
+    void streaming() throws Exception {
+        StreamingQuery stream = newSparkSession().readStream()
+            .schema(new StructType()
```
rjrudin, line 44: Have to provide the schema upfront.

src/test/java/com/marklogic/spark/writer/WriteRowsTest.java
anu3990: Spark is expecting a directory here when stream is used.

PR #25, opened 2023-05-09
DEVEXP-430 : Write stream of rows

anu3990:


src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+        StreamingQuery query = dataset.writeStream()
+            .outputMode("append") // TODO: Investigate for complete and update modes
+            .format(Options.FORMAT)
+            .option(Options.HOST, testConfig.getHost())
+            .option(Options.PORT, testConfig.getRestPort())
+            .option(Options.USERNAME, testConfig.getUsername())
+            .option(Options.PASSWORD, testConfig.getPassword())
+            .option(Options.WRITE_COLLECTIONS, "my-test-data")
+            .option(Options.CheckpointLocation,"src/test/resources/checkpoint/"+
```
rjrudin, line 54: This results in non-gitignored files being added to the project. I think TempDir is a good fit here - just add this as an argument to the method:

    void streamRowsFromCsvFile(@TempDir Path tempDir)

And then you can do this in the option:

    .option(Options.CheckpointLocation, tempDir.toFile().getAbsolutePath())

We don't care about the contents of that directory at all, at least not in the scope of this test. So it's fine for it to be a temp directory, which also gets deleted by default after the test ends.

src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .option("header", true)
+            .format("csv")
+            .load("src/test/resources/inputForStream"); // Needs to be a directory when using streams
+
+        StreamingQuery query = dataset.writeStream()
+            .outputMode("append") // TODO: Investigate for complete and update modes
+            .format(Options.FORMAT)
+            .option(Options.HOST, testConfig.getHost())
+            .option(Options.PORT, testConfig.getRestPort())
+            .option(Options.USERNAME, testConfig.getUsername())
```
rjrudin, line 51: This uses the values of mlUsername/mlPassword which is expected to be an admin-like user; we want a more restricted user for these tests. So use what the other tests are using:

    .option("spark.marklogic.client.username", "spark-test-user")
    .option("spark.marklogic.client.password", "spark")

src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .add("Name", DataTypes.StringType)
+            .add("House", DataTypes.StringType);
+
+        Dataset<Row> dataset = sparkSession.readStream().schema(struct)
+            .option("header", true)
+            .format("csv")
+            .load("src/test/resources/inputForStream"); // Needs to be a directory when using streams
+
+        StreamingQuery query = dataset.writeStream()
+            .outputMode("append") // TODO: Investigate for complete and update modes
```
rjrudin, line 47: I actually just opened a JIRA task for this - 445.

src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.spark.sql.streaming.StreamingQuery;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructType;
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.TimeoutException;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class StreamRowsTest extends AbstractIntegrationTest {
```
rjrudin, line 33: Can you add a test for when the user has invalid config? I'm curious what happens to the error - we can at least get that documented via a test. I'm assuming it will be thrown by `processAllAvailable` but I don't know. 

We want an error that won't occur until the connector tries to write data, so setting `spark.marklogic.write.transformName` to an invalid REST transform name is a good way to achieve that.

rjrudin (changes requested):
Looks good, mostly comments on improving the test.

src/main/java/com/marklogic/spark/writer/MarkLogicWrite.java
```diff
@@ -53,4 +58,19 @@ public void abort(WriterCommitMessage[] messages) {
...
         }
     }
+
+    @Override
+    public StreamingDataWriterFactory createStreamingWriterFactory(PhysicalWriteInfo info) {
+        return new MarkLogicDataWriterFactory(writeContext);
+    }
+
+    @Override
+    public void commit(long epochId, WriterCommitMessage[] messages) {
```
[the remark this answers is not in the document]
anu3990: Added logging for epochId. It will be logged even if the messages are empty.

src/main/java/com/marklogic/spark/writer/MarkLogicWrite.java
```diff
@@ -53,4 +58,25 @@ public void abort(WriterCommitMessage[] messages) {
...
+        logger.info("EpochId for commit is "+ epochId);
```
rjrudin, line 69: Small thing, which I'll tweak later - `logger` functions allow for token replacement, so you can do `logger.info("epochId for commit is {}", epochId)` . I keep trying to remember that, I know Sonar (which we don't yet have running) will complain about it otherwise.

rjrudin (approved):

src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Test
+    void inValidTransform(@TempDir Path tempDir) {
+        try{
+            newDefaultStreamWriter(newSparkSession(), "src/test/resources/inputForStream", struct, "csv")
+                .outputMode("append")
+                .option(Options.WRITE_COLLECTIONS, "my-test-data")
+                .option(Options.WRITE_TRANSFORM_NAME, "this-doesnt-exist")
+                .option("checkpointLocation", tempDir.toFile().getAbsolutePath()) // checkpointLocation needs to be a new location everytime job is run else docs are not written.
+                .start().processAllAvailable();
+        } catch (Exception ex){
```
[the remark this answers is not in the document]
rjrudin: Coincidentally, we have an error on the CSV test in Jenkins - http://jenkins1.marklogic.com:9000/blue/organizations/jenkins/marklogic-spark-connector/detail/PR-25/7/tests - and the `assertTrue(doc.has("House"));` call is failing - but we don't know why. That's where something like `assertTrue(doc.has("House"), "Unexpected JSON: " + doc)` would help as we'd be able to see what the doc looks like.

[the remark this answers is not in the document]
rjrudin: My bet is that since this is grabbing a random JSON doc from the collection, it's getting the latest one that has a name but no house. I'll make the test 100% reliable and a little more fun by using the new uriTemplate feature, and then the test can ask for a specific URI that it knows has both columns.

PR #31, opened 2023-05-11
DEVEXP-430 Fix for streaming test

rjrudin:
Tweaked logging a bit too, and also removed Serializable from MarkLogicWrite - I had added that in my prototype, but I think I did that based on something wrong I was doing. It does not appear needed, which makes sense - the point of the Write object is to produce a serializable Factory that can be sent to a different Spark host.


src/test/java/com/marklogic/spark/writer/StreamRowsTest.java
```diff
@@ -26,39 +30,63 @@
...
+
+        // No error will be thrown yet as the streaming occurs in the background, though we'll likely see errors
+        // from the background thread. Calling processAllAvailable should force an error to occur.
+
+        StreamingQueryException ex = assertThrows(StreamingQueryException.class, () -> query.processAllAvailable());
+        assertTrue(ex.getMessage().contains("Extension this-doesnt-exist or a dependency does not exist"),
+            "Unexpected error: " + ex.getCause());
     }
+
+    private DataStreamWriter newDefaultStreamWriter(SparkSession session, String path, StructType struct, String format) {
```
rjrudin, line 77: I moved this here, as it's only used by this test. If we have a separate test class in the future that needs this, it can move up to the parent class. But until then, having it here makes this test easier to read and understand.

src/main/java/com/marklogic/spark/writer/MarkLogicWrite.java
```diff
@@ -66,17 +63,15 @@ public StreamingDataWriterFactory createStreamingWriterFactory(PhysicalWriteInfo
 
     @Override
     public void commit(long epochId, WriterCommitMessage[] messages) {
-        logger.info("EpochId for commit is "+ epochId);
         if (messages != null && messages.length > 0) {
-            logger.info("Commit messages received: {}", Arrays.asList(messages));
+            logger.info("Commit messages received for epochId {}: {}", epochId, Arrays.asList(messages));
```
rjrudin, line 67: Combined this into one log message; one log event is typically preferable for tools like Grafana and other log analyzers.

BillFarber (approved):

src/main/java/com/marklogic/spark/writer/MarkLogicWrite.java
anu3990: We might need to log epochId, even when the commit messages are null or empty.

anu3990 (changes requested):

src/main/java/com/marklogic/spark/writer/MarkLogicWrite.java
rjrudin: I changed it to always log epochId and messages. I'm going to do some testing to see when messages are null, as I've encountered that before. Not sure if it matters to log messages when it's null, that's probably fine.

anu3990 (approved):

PR #32, opened 2023-05-11
DEVEXP-451 Prototype of readStream support

rjrudin:


src/test/java/com/marklogic/spark/reader/ReadStreamOfRowsTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.marklogic.spark.AbstractIntegrationTest;
+import com.marklogic.spark.Options;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.nio.file.Path;
+
+public class ReadStreamOfRowsTest extends AbstractIntegrationTest {
+
```
BillFarber, line 11: I don't understand what makes this test a micro batch?
I would have expected to see an option that specifies micro batch.

rjrudin: It's the use of `readStream` instead of `read`. That's what took me awhile to figure out - but to implement `readStream`, a connector either implements "micro batch" or "continuous" or both. I still have to figure out what "continuous" means and if that's worth implementing too.

PR #33, opened 2023-05-15
DEVEXP-451 Can now use readStream

rjrudin:
Most of the new stuff is in the new `MarkLogicMicroBatchStream` class, which uses a bucket index as an offset.

### DIFF
--- a/src/main/java/com/marklogic/spark/ContextSupport.java
+++ b/src/main/java/com/marklogic/spark/ContextSupport.java
@@ -20,7 +20,12 @@ public class ContextSupport implements Serializable {
 
     public DatabaseClient connectToMarkLogic() {
         Map<String, String> connectionProps = buildConnectionProperties();
-        DatabaseClient client = DatabaseClientFactory.newClient(propertyName -> connectionProps.get("spark." + propertyName));
+        DatabaseClient client;
+        try {
+            client = DatabaseClientFactory.newClient(propertyName -> connectionProps.get("spark." + propertyName));
+        } catch (Exception e) {
+            throw new RuntimeException(String.format("Unable to connect to MarkLogic; cause: %s", e.getMessage()), e);
+        }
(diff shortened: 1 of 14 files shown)


src/main/java/com/marklogic/spark/ContextSupport.java
```diff
@@ -20,7 +20,12 @@ protected ContextSupport(Map<String, String> properties) {
 
     public DatabaseClient connectToMarkLogic() {
         Map<String, String> connectionProps = buildConnectionProperties();
-        DatabaseClient client = DatabaseClientFactory.newClient(propertyName -> connectionProps.get("spark." + propertyName));
+        DatabaseClient client;
+        try {
+            client = DatabaseClientFactory.newClient(propertyName -> connectionProps.get("spark." + propertyName));
+        } catch (Exception e) {
+            throw new RuntimeException(String.format("Unable to connect to MarkLogic; cause: %s", e.getMessage()), e);
```
rjrudin, line 27: Added this for better context when a connection cannot be made.

src/main/java/com/marklogic/spark/reader/ReadContext.java
```diff
@@ -69,7 +69,9 @@ public ReadContext(Map<String, String> properties, boolean inferSchema) {
             SparkSession.active().sparkContext().defaultMinPartitions(), 1);
         final long batchSize = getNumericOption(ReadConstants.BATCH_SIZE, DEFAULT_BATCH_SIZE, 0);
         final String dslQuery = properties.get(ReadConstants.OPTIC_DSL);
-
+        if (dslQuery == null || dslQuery.trim().length() < 1) {
```
rjrudin, line 72: Added this as I forgot the query in one of my ad hoc tests, and the error didn't indicate what the problem was

BillFarber (dismissed):

anu3990 (dismissed):

anu3990 (approved):

PR #38, opened 2023-05-19
DEVEXP-461 Now pushing down filters

rjrudin:


src/main/java/com/marklogic/spark/reader/ReadContext.java
```diff
@@ -109,23 +115,54 @@ Iterator<JsonNode> readRowsInBucket(RowManager rowManager, PlanAnalysis.Partitio
...
+        // TODO Change this to use resultRows once Java Client 6.2.1 is available. That should perform better as it
+        // avoids creating a JsonNode.
+        JsonNode result = rowManager.resultDoc(plan, jsonHandle).get();
         return result != null && result.has("rows") ?
             result.get("rows").iterator() :
             new ArrayList<JsonNode>().iterator();
     }
 
-    public StructType getSchema() {
+    private PlanBuilder.Plan buildPlanForBucket(RowManager rowManager, PlanAnalysis.Bucket bucket) {
```
rjrudin, line 130: I extracted this method to make the calling method a bit simpler.

src/main/java/com/marklogic/spark/reader/filter/FilterFactory.java
```diff
@@ -0,0 +1,59 @@
+package com.marklogic.spark.reader.filter;
+
+import org.apache.parquet.filter2.predicate.Operators;
+import org.apache.spark.sql.sources.*;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class FilterFactory {
```
rjrudin, line 9: This is the thing that knows how to convert a Spark Filter into an "OpticFilter", which then knows how to modify the user's serialized Optic query and bind a param value if necessary.

BillFarber (approved):
You're right. This is pretty elegant.

anu3990 (approved):

PR #39, opened 2023-05-22
DEVEXP-462 Now pushing down orderBy, limit, and offset

rjrudin:
Realized that orderBy needs to be done with limit per Spark's `SupportsPushDownTopN` interface. 

All 3 are implemented simply by adding the appropriate operator to the user's serialized Optic plan.


BillFarber (approved):
LGTM, but the logic is getting really complex. Will Spark users understand all this?

PR #40, opened 2023-05-23
Fixing rowid handling on MarkLogic 10

rjrudin:


src/test/java/com/marklogic/spark/AbstractIntegrationTest.java
```diff
@@ -93,4 +95,13 @@ protected String readClasspathFile(String path) {
     protected final String makeClientUri() {
         return String.format("%s:%s@%s:%d", TEST_USERNAME, TEST_PASSWORD, testConfig.getHost(), testConfig.getRestPort());
     }
+
+    protected final boolean isMarkLogic10() {
```
BillFarber, line 99: Should this be in the Java client?

rjrudin: See the comment in the MarkLogicVersion class - "This is duplicated from java-client-api and could be a good addition to marklogic-client-api.". I'm thinking it's worth adding there, possibly in 6.3.0, as this is duplicated in a few places now.

PR #47, opened 2023-05-26
Fixes for count/orderBy on qualified column names

rjrudin:
And... I implemented groupBy + count while I was in here fixing things as it was trivial to do. Also tweaked some logging based on testing. But the main thing is the addition of tests to verify that column names work, regardless of whether they have no qualifier, only a view, or a schema + view.

### DIFF
--- a/src/main/java/com/marklogic/spark/reader/PlanUtil.java
+++ b/src/main/java/com/marklogic/spark/reader/PlanUtil.java
@@ -4,10 +4,14 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.marklogic.spark.reader.filter.OpticFilter;
+import org.apache.spark.sql.connector.expressions.Expression;
+import org.apache.spark.sql.connector.expressions.NamedReference;
 import org.apache.spark.sql.connector.expressions.SortDirection;
 import org.apache.spark.sql.connector.expressions.SortOrder;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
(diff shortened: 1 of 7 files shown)


src/main/java/com/marklogic/spark/reader/PlanUtil.java
```diff
@@ -37,30 +50,41 @@ static ObjectNode buildOffset(int offset) {
...
         final String direction = SortDirection.ASCENDING.equals(sortOrder.direction()) ? "asc" : "desc";
-        final String columnName = sortOrder.expression().describe();
-        return newOperation("order-by", args -> args.addObject()
-            .put("ns", "op").put("fn", direction)
-            .putArray("args").addObject()
-            .put("ns", "op").put("fn", "col").putArray("args").add(columnName));
+        final String columnName = expressionToColumnName(sortOrder.expression());
+        return newOperation("order-by", args -> {
+            ArrayNode orderByArgs = args.addObject().put("ns", "op").put("fn", direction).putArray("args");
+            // This may be a bad hack to account for when the user does a groupBy/count/orderBy/limit, which does not
```
rjrudin, line 56: This is weird, but it's covered by multiple tests.

src/main/java/com/marklogic/spark/reader/PlanUtil.java
```diff
@@ -37,30 +50,41 @@ static ObjectNode buildOffset(int offset) {
...
-                    colArgs.add(objectMapper.nullNode()).add(parts[0]).add(parts[1]);
-                } else {
-                    colArgs.add(objectMapper.nullNode()).add(objectMapper.nullNode()).add(parts[0]);
-                }
+                populateSchemaCol(innerArgs.addObject(), field.name());
             }
         });
     }
 
+    private static void populateSchemaCol(ObjectNode node, String columnName) {
```
rjrudin, line 76: Added this and using it any time we have to refer to a column. I still need to update the filter tests to ensure they're using this as well.

BillFarber (approved):
LGTM

PR #60, opened 2023-06-06
Fix for not pushing down when count of rows is zero

rjrudin:
Need to do this in the ScanBuilder so we can return "false" to Spark to let it know that we did not push down the operation. 

Also added an assertion to WriteRowsTest to help debug what seems like an intermittent failure on Jenkins.


BillFarber (approved):
LGTM

PR #69, opened 2023-06-08
Improving groupBy pushdown to work for multiple column names

rjrudin:
Turns out this was a simple enhancement since `op.groupBy` already supports multiple column names being passed in.

### DIFF
--- a/src/main/java/com/marklogic/spark/reader/MarkLogicScanBuilder.java
+++ b/src/main/java/com/marklogic/spark/reader/MarkLogicScanBuilder.java
@@ -145,11 +145,10 @@ public class MarkLogicScanBuilder implements ScanBuilder, SupportsPushDownFilter
         }
         if (supportCompletePushDown(aggregation)) {
             if (aggregation.groupByExpressions().length > 0) {
-                Expression expr = aggregation.groupByExpressions()[0];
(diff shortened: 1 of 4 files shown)


src/main/java/com/marklogic/spark/reader/MarkLogicScanBuilder.java
```diff
@@ -167,12 +166,10 @@ public boolean supportCompletePushDown(Aggregation aggregation) {
             return false;
         }
         AggregateFunc[] expressions = aggregation.aggregateExpressions();
-        if (expressions.length == 1 && expressions[0] instanceof CountStar) {
-            // If a count() is used, it's supported if there's no groupBy - i.e. just doing a count() by itself -
-            // and supported with a single groupBy - e.g. groupBy("column").count().
-            return aggregation.groupByExpressions().length < 2;
-        }
-        return false;
+        // If a count() is used, it's supported if there's no groupBy - i.e. just doing a count() by itself -
```
rjrudin, line 169: I think I made this needlessly complicated before - since `op.groupBy` supports multiple column names, we don't have to care how many column names the user passed into the Spark `groupBy` call.

BillFarber (approved):

PR #87, opened 2023-10-18
Added docs for reading via custom queries

rjrudin:
- Added a TOC
- Moved all the Optic-specific sections under the overall "Reading via Optic" section
- Bumped "Error Handling" to the bottom since it'll be the same for both Optic and custom queries

There are some whitespace changes in terms of newline symbols, so be sure to ignore those when reviewing in GH.


BillFarber:
Seems like a good direction.

PR #92, opened 2023-10-30
DEVEXP-616 Added entity aggregation example project

rjrudin:


examples/entity-aggregation/src/main/java/org/example/ImportCustomers.java
```diff
@@ -0,0 +1,70 @@
...
+        String query = "(select c.*, r.rental_id, r.rental_date, r.return_date " +
+            "from customer c " +
+            "inner join public.rental r on c.customer_id = r.customer_id " +
+            ") the_query ";
+
+        session
+            .read()
+            .jdbc("jdbc:postgresql://localhost/dvdrental", query, jdbcProperties)
+            // Group all the rows by customer_id.
+            .groupBy("customer_id")
```
BillFarber, line 43: The following agg function call is complex and it's where the important magic is happening.
I wonder if there should be something in here pointing to the example JSON in the README file and specifically telling users that this code is transforming multiple rows into single rows that (eventually) look like the JSON there.

BillFarber (dismissed):

BillFarber (approved):

PR #107, opened 2023-11-13
DEVEXP-627 Can now read via user-defined partitions

rjrudin:
This replaces the "get batch ids" concept used for streaming, such that both regular reads and streaming reads now use the optional "get partitions" feature.

Did some doc updates, but some sections need to be rewritten now. Will do that in a follow up PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ gradle-local.properties
 logs
 .ipynb_checkpoints
 venv
+.venv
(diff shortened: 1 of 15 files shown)


rjrudin:
I'm working on a follow up PR now to modifying the docs to make it clear that user-defined partitions work for regular/streaming reads.

BillFarber (approved):

PR #117, opened 2023-12-07
Fixing all Sonar issues

rjrudin:
All of these are just minor things. I did introduce `ConnectorException` to avoid the use of generic exceptions, but that extends `RuntimeException` itself. So no functional changes here, just making Sonar happy.


BillFarber (approved):
LGTM

PR #131, opened 2023-12-23
DEVEXP-673 Can now read zip files

rjrudin:
Main thing of note here is that we extend the Spark `FileTable` class, which is extended by all of Spark's OOTB file-based data sources. This is the only way I could find to reuse the super-valuable code for creating a file index, which applies all of Spark's support for finding files based on one or more paths. 

And since `MarkLogicFileTable` is a `Table`, this becomes a new batch of reader components that branches off from `DefaultSource`. We will likely do the same thing when we support "Read documents from MarkLogic" instead of trying to shoehorn that into the current Optic-based reader.


src/main/java/com/marklogic/spark/DefaultSource.java
```diff
@@ -100,11 +102,37 @@ public boolean supportsExternalMetadata() {
...
+
     private boolean isReadOperation(Map<String, String> properties) {
         return properties.get(Options.READ_OPTIC_QUERY) != null || isReadWithCustomCodeOperation(properties);
     }
 
     private boolean isReadWithCustomCodeOperation(Map<String, String> properties) {
         return Util.hasOption(properties, Options.READ_INVOKE, Options.READ_XQUERY, Options.READ_JAVASCRIPT);
     }
+
+    private StructType inferSchemaFromOpticQuery(Map<String, String> caseSensitiveOptions) {
```
rjrudin, line 117: No new code here, I just moved this out of the `inferSchema` method to simplify that method.

src/main/java/com/marklogic/spark/reader/file/MarkLogicFileTable.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class MarkLogicFileTable extends FileTable {
+
+    public MarkLogicFileTable(SparkSession sparkSession, CaseInsensitiveStringMap options, Seq<String> paths) {
+        super(sparkSession, options, paths, Option.apply(FileRowSchema.SCHEMA));
+    }
+
+    @Override
+    public ScanBuilder newScanBuilder(CaseInsensitiveStringMap options) {
+        return new FileScanBuilder(super.fileIndex());
```
rjrudin, line 35: This is the magic method - fileIndex.

src/main/java/com/marklogic/spark/reader/file/ZipFileReader.java
```diff
@@ -0,0 +1,94 @@
...
+    private ZipEntry currentZipEntry;
+
+    ZipFileReader(FilePartition partition, SerializableConfiguration hadoopConfiguration) {
+        this.path = partition.getPath();
+        try {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Reading zip file {}", this.path);
+            }
+            Path hadoopPath = new Path(this.path);
+            this.zipInputStream = new ZipInputStream(hadoopPath.getFileSystem(hadoopConfiguration.value()).open(hadoopPath));
```
rjrudin, line 34: This is a critical bit here too - MLCP does the same thing for resolving a string path into a local/HDFS/S3 path.

BillFarber (approved):

PR #137, opened 2024-01-03
4115: Can now read aggregate XML files

rjrudin:


BillFarber (approved):

rjrudin:
Added "XML" to the reader class.

BillFarber (approved):

PR #150, opened 2024-01-31
MLE-3943: Updating the pipeline to use marklogic docker

SameeraPriyathamTadikonda:


Jenkinsfile
```diff
@@ -53,7 +51,13 @@ pipeline{
       }
       agent {label 'devExpLinuxPool'}
       steps{
-        runtests('Latest','11','JAVA11_HOME_DIR')
+        sh label:'mlsetup', script: '''#!/bin/bash
+            sudo /usr/local/sbin/mladmin remove
```
rjrudin, line 55: What do these two lines do? I'm thinking that since they're repeated below, it'd be good to put them into a function that has a comment on it explaining the purpose of these lines.

SameeraPriyathamTadikonda: Added a print statement.

Line 54 - Opens a shell terminal and 55 removes any existing marklogic.

Jenkinsfile
```diff
@@ -1,14 +1,13 @@
...
-  copyRPM mlVersionType,mlVersion
-  setUpML '$WORKSPACE/xdmp/src/Mark*.rpm'
+def runtests(String javaVersion){
   sh label:'test', script: '''#!/bin/bash
     export JAVA_HOME=$'''+javaVersion+'''
     export GRADLE_USER_HOME=$WORKSPACE/$GRADLE_DIR
     export PATH=$GRADLE_USER_HOME:$JAVA_HOME/bin:$PATH
     cd marklogic-spark-connector
     echo "mlPassword=admin" > gradle-local.properties
+    sleep 30s
```
rjrudin, line 10: Why is this here? If it's needed, add a comment before it to explain why it's sleeping and why 30s is the chosen value.

Jenkinsfile
```diff
@@ -53,11 +52,29 @@ pipeline{
       }
       agent {label 'devExpLinuxPool'}
       steps{
-        runtests('Latest','11','JAVA11_HOME_DIR')
+        sh label:'mlsetup', script: '''#!/bin/bash
+            echo "Removing any running marklogic server and clean up marklogic data directory"
```
rjrudin, line 56: Small nit - always do "MarkLogic" in a comment/printout, not "marklogic".

rjrudin (changes requested):

Jenkinsfile
SameeraPriyathamTadikonda: Done

Jenkinsfile
SameeraPriyathamTadikonda: Added a comment.

rjrudin (approved):

PR #151, opened 2024-01-31
Initial docs for reading documents

rjrudin:
These aren't intended to be finished, they're just a starting point for some consultants to try out. 

Also fixed an issue where batch size wasn't being applied. Bumped to the latest Java Client to fix a bug with retrieving "metadata values" too.


docs/reading.md
rjrudin, line 1: I broke this out into multiple child pages, as this page was becoming very long.

docs/reading-data/custom-code.md
```diff
@@ -0,0 +1,184 @@
+---
```
rjrudin, line 1: None of this is new, it's just broken out from the parent page.

docs/reading-data/optic.md
```diff
@@ -0,0 +1,280 @@
+---
```
rjrudin, line 1: Nothing new, just broken out from parent page.

anu3990 (approved):

PR #157, opened 2024-02-06
MLE-12218 User can now configure partitions per forest

rjrudin:
The main guts here is in the new `ForestPartitionPlanner` along with changes to `UrisBatcher`. 

Interestingly, had to add "queryFormat" as a parameter in order to call `queryManager.search`, whereas `queryManager.uris` doesn't require it. 

Updated a few tests based on a slight tweak to how invalid number values are handled.


BillFarber (dismissed):
Not sure if I should approve this given the test failures.

BillFarber (approved):

PR #162, opened 2024-02-12
MLE-12296 Added an example of nested aggregation.

BillFarber:


examples/entity-aggregation/src/main/java/org/example/ImportCustomersWithRentalsAndPayments.java
```diff
@@ -0,0 +1,71 @@
...
+        // Define the JDBC properties for connecting to Postgres.
+        Properties jdbcProperties = new Properties();
+        jdbcProperties.setProperty("driver", "org.postgresql.Driver");
+        jdbcProperties.setProperty("user", "postgres");
+        jdbcProperties.setProperty("password", "postgres");
+
+        // See https://stackoverflow.com/a/39129546 for information on aliased queries with Spark. A table name may be
+        // used, but in this example, we want to join customers, rentals, and payments together. The query is placed in
+        // parentheses with an alias after it; the alias can have any name.
+        String query =
```
rjrudin, line 31: Take a look at `ImportJdbcCommand` in our ETL project - I am thinking that is the better, simpler way to use JDBC with Spark so this complex "aliased query" isn't needed (this was the first approach I found with Spark JDBC so I just went with it), which means the comment can go away too. We'd instead use the "query" option which seems a lot simpler. If you agree, update this and the other example program too.

examples/entity-aggregation/README.md
```diff
@@ -115,3 +115,82 @@ Note as well that this example is not intended to be authoritative. Please see
...
+"customers" table, so you will see 10 customer documents in the Documents database.
+
+As an example, one of these customers (#182), has a single rental (#4591) that has multiple payments. If you examine the
+document for that customer, (/customerWithDoubleNesting/182.json) you will be able to verify this nested aggregation:
+
+```
+{
+  "customer_id": 182,
+  "last_name": "Lane",
+  "Rentals w/ Payments": [
```
rjrudin, line 147: Just call this "rentals", that's more realistic as I don't think it's common to ever have a "/" in a JSON field name.

examples/entity-aggregation/README.md
```diff
@@ -115,3 +115,82 @@ Note as well that this example is not intended to be authoritative. Please see
 on writing Spark programs. You may also find
 [this reference on Spark aggregate functions](https://sparkbyexamples.com/spark/spark-sql-aggregate-functions/) helpful.
 
+## Importing customers, rentals and payments from Postgres to MarkLogic
```
rjrudin, line 118: I think this should not duplicate most of the text above it, but rather explain what it shows in addition to the first example. I'm envisioning something like "This project also contains an example of two levels of joins being aggregated together" (I'm not sure of the correct terminology to use here). "In this example, rentals are aggregated into customers, and payments are also aggregated into rentals, producing the following customer documents (show JSON next). To try this out, run the following Gradle command..."

Basically, you can assume the user has read the stuff above it so we don't need to duplicate so much of the instructions again. 

Alternatively, leave the text as-is but add a new ticket to the 2.2.0 Spark connector release to clean up the README so we don't have so much duplicated text.

rjrudin (changes requested):
Looks good, just one small tweak. For the other two ideas, can either try them as part of this PR or toss them into a new JIRA ticket for later.

rjrudin (approved):

PR #166, opened 2024-02-16
MLE-12332 Can now read RDF/XML files

rjrudin:
Main things of note:

1. New `TripleRowSchema` instead of using document row schema, as we'll need to capture the graph for quads.
2. `TripleSerializer` is mostly copy/pasted (and cleaned up a bunch) from MLCP.

### DIFF
--- a/src/main/java/com/marklogic/spark/reader/file/TripleRowSchema.java
+++ b/src/main/java/com/marklogic/spark/reader/file/TripleRowSchema.java
@@ -1,0 +1,22 @@
+package com.marklogic.spark.reader.file;
+
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructType;
+
+/**
+ * Represents a triple as read from an RDF file and serialized into the 3 XML elements comprising
+ * a MarkLogic triple.
+ */
+public abstract class TripleRowSchema {
+
+    public static final StructType SCHEMA = new StructType()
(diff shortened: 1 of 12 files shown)


src/main/java/com/marklogic/spark/reader/file/TripleRowSchema.java
```diff
@@ -0,0 +1,22 @@
...
+ * Represents a triple as read from an RDF file and serialized into the 3 XML elements comprising
+ * a MarkLogic triple.
+ */
+public abstract class TripleRowSchema {
+
+    public static final StructType SCHEMA = new StructType()
+        .add("subject", DataTypes.StringType, false)
+        .add("predicate", DataTypes.StringType, false)
+        .add("object", DataTypes.StringType, false)
+        .add("datatype", DataTypes.StringType)
```
BillFarber, line 16: I don't know enough about triples - can each part of the triple have a different datatype?

BillFarber (approved):

src/main/java/com/marklogic/spark/reader/file/TripleRowSchema.java
rjrudin: Only the object - subject/predicate are always IRIs and thus always strings.

rjrudin:
Merging this - Jenkins is failing to start Docker, will debug separately.

PR #172, opened 2024-02-21
Added mention of a load balancer to the docs

rjrudin:


wattsferry (approved):
I think this is good. About the only other thing you might add (not required) is that not all load balancers retry automatically, but that if they do, they should be set to retry until at least 1 sec greater than the MarkLogic app server default timeout (30 seconds).

BillFarber (approved):

PR #181, opened 2024-02-29
MLE-12386 : Read MLCP archive file

anu3990:


src/test/java/com/marklogic/spark/reader/file/MlcpArchiveFileReaderTest.java
```diff
@@ -0,0 +1,44 @@
...
+import com.marklogic.spark.Options;
+import org.apache.spark.sql.DataFrameReader;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashSet;
+import java.util.List;
+
+public class MlcpArchiveFileReaderTest extends AbstractIntegrationTest {
```
rjrudin, line 13: For consistency with the other tests, name this `ReadMlcpArchiveFilesTest`. 

Also, Sonar is oddly picky about test classes being `public` so remove that keyword as well. You can run `./gradlew sonar` to see the Sonar report, as long as you have Sonar running via Docker.

rjrudin (approved):
Good start - just rename the test and then merge away.

The next story - reading an MLCP archive with metadata - is probably not worth doing yet since we'll want to write an XSLT to transform the MLCP metadata into REST API metadata. I think MLE-12387 will be better to do next. I'll add a sample of what a REST API metadata file looks like to that story.

rjrudin (approved):

PR #184, opened 2024-03-05
MLE-12717 Can now convert MLCP metadata into REST metadata

rjrudin:
This is intended to be used by MLE-12388 - importing an MLCP archive. The intent is that an MLCP metadata document can be converted by `MlcpMetadataConverter` into a `DocumentMetadataHandle` object, which can then be converted into the appropriate column values in our `DocumentRowSchema`.


BillFarber (commented):
Looks good, but WriteDocumentZipFilesTest tests are failing - presumably the changes to the query changed the number of results.

rjrudin:
> Looks good, but WriteDocumentZipFilesTest tests are failing - presumably the changes to the query changed the number of results.

Oops, that was an accidental check-in.

BillFarber (approved):

PR #187, opened 2024-03-06
MLE-12627 : Write archive files

anu3990:


src/test/java/com/marklogic/spark/writer/file/WriteDocumentZipFilesTest.java
```diff
@@ -132,4 +160,30 @@ private void verifyZipFilesContainFifteenAuthors(Path tempDir) throws IOExceptio
...
+            assertTrue(uri.startsWith(expectedUriPrefix), "Unexpected URI, which is expected to start with the " +
+                "absolute path of the zip file: " + uri);
+
+            if(uri.endsWith(".json")) {
+                JsonNode doc = objectMapper.readTree((byte[]) row.get(3));
+                assertTrue(doc.has("CitationID"), "Unexpected JSON: " + doc);
+            } else {
+                assertTrue(uri.endsWith(".metadata"));
+                String metadataContent = new String((byte[]) row.get(3), StandardCharsets.UTF_8);
+                assertTrue(metadataContent.contains("<rapi:collection>author</rapi:collection>"));
```
rjrudin, line 185: We generally want to avoid any sort of "contains" assertion on XML. The "rapi" prefix isn't guaranteed to exist - it could be the default namespace of the XML and the prefix wouldn't exist, in which case this assertion would fail. 

Instead, use `XmlNode` which is a convenient test class from marklogic-junit5:

```
XmlNode metadata = new XmlNode(metadataContent, Namespace.getNamespace("rapi", "http://marklogic.com/rest-api"));
                metadata.assertElementValue("/rapi:metadata/rapi:collections/rapi:collection", "author");
```

That `assertElementValue` call will work even if the "rapi" prefix isn't present.

rjrudin (changes requested):
Looks good, just a few small things.

rjrudin:
@anu3990  I'm going to merge this and then update #186 to make some helpful refactoring improvements. Can you do a PR after that refactoring to address the 3 comments I made in this?

rjrudin (approved):

PR #190, opened 2024-03-07
MLE-12627 : Write archive files

anu3990:


src/main/java/com/marklogic/spark/writer/file/ZipFileWriter.java
```diff
@@ -57,8 +57,8 @@ public void write(InternalRow row) throws IOException {
         zipOutputStream.putNextEntry(new ZipEntry(entryName));
         this.contentWriter.writeContent(row, zipOutputStream);
         zipEntryCounter++;
-        if(properties.get(Options.READ_DOCUMENTS_CATEGORIES) != null &&
-            properties.get(Options.READ_DOCUMENTS_CATEGORIES).contains("metadata")){
+        if(!row.isNullAt(3) || (properties.get(Options.READ_DOCUMENTS_CATEGORIES) != null &&
```
rjrudin, line 60: This should ignore the properties - just because they exist doesn't mean the rows have any metadata (also, just checking for "metadata" wouldn't suffice - the user could do "content,collections"). We instead want:

```
if (!row.isNullAt(3) || !row.isNullAt(4) || !row.isNullAt(5) || 
    !row.isNullAt(6) || !row.isNullAt(7)) {
    // add metadata entry
}
```

That also means that we'd want to test each of those conditions. So you should have 5 separate tests - one where collections is populated, then permissions, then quality, then properties, then metadata values. Those tests don't necessarily need to examine the ".metadata" file, they just need to verify that the ".metadata" file exists.

rjrudin (changes requested):

src/main/java/com/marklogic/spark/writer/file/ZipFileWriter.java
anu3990: User can have Options.READ_DOCUMENTS_CATEGORIES = "content,collections" as well right.

rjrudin: Right, but the options aren't reliable as an indicator. We'll use the `READ_DOCUMENTS_CATEGORIES` option in our tests to make life easy. But a user could manually construct a Spark dataframe and populate the metadata column manually, in which case the `READ_DOCUMENTS_CATEGORIES` option won't exist. 

This is all due to the reader and writer being decoupled in Spark. The writer can't assume anything about how the data was read. So it can't assume that `READ_DOCUMENTS_CATEGORIES` was used. All it can do is look at the columns in the row.

rjrudin (dismissed):
Looks good! Note that if you run `./gradlew sonar`, you'll get 4 smells. Two of them are easy to fix - they're unused imports. A third is easy - make the test class non-public.

The fourth is more interesting and you may enjoy refactoring it, though you can do that in a separate PR. It's converting the 6 tests in `WriteArchiveTest` into a single parameterized test. I think `CsvSource` would work well for it. Search on "ParameterizedTest" in src/test/java in this repo for other examples.

rjrudin (approved):

PR #201, opened 2024-03-19
MLE-12387 : Read NT archive file

anu3990:


src/test/java/com/marklogic/spark/reader/file/ReadArchiveFileTest.java
```diff
@@ -0,0 +1,41 @@
...
+        hashSet.add("/betsy");
+        hashSet.add("/george");
+        hashSet.add("/martha");
+        hashSet.add("/mary");
+
+        for(int i =0; i<5; i++){
+            String temp = rows.get(i).getString(0).substring(rows.get(i).getString(0).lastIndexOf("/"));
+            System.out.println("****************");
+            System.out.println(new String((byte[]) rows.get(i).get(3), "UTF-8"));
+            assert(hashSet.contains(temp));
```
anu3990, line 36: This is working and I can also see the doc content along with the corresponding metadata content. The only problem is - the content comes after the metadata content. For example, the first row  has the content for alexander.metadata followed by content in document alexander (from sample-XML.zip)

src/test/java/com/marklogic/spark/reader/file/ReadArchiveFileTest.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+public class ReadArchiveFileTest extends AbstractIntegrationTest {
+
+    @Test
+    void readArchiveFile() throws UnsupportedEncodingException {
+        Dataset<Row> reader = newSparkSession().read()
+            .format(CONNECTOR_IDENTIFIER)
+            .option(Options.READ_FILES_TYPE, "archive")
+            .load("src/test/resources/mlcp-archive-files/sample-XML.zip");
```
rjrudin, line 20: This is loading an MLCP archive, but for MLE-12387, we want to read an NT archive file. Best way to do that is to reuse the test code from `WriteArchiveTest`. Can just copy and paste from there for now - e.g. I did a quick test like this:

```
@BeforeEach
    void beforeEach() {
        TestUtil.insertTwoDocumentsWithAllMetadata(getDatabaseClient());
    }

    @Test
    void test(@TempDir Path tempDir) {
        newSparkSession().read().format(CONNECTOR_IDENTIFIER)
            .option(Options.CLIENT_URI, makeClientUri())
            .option(Options.READ_DOCUMENTS_COLLECTIONS, "collection1")
            .option(Options.READ_DOCUMENTS_CATEGORIES, "content,metadata")
            .load()
            .repartition(1)
            .write().format(CONNECTOR_IDENTIFIER)
            .option(Options.WRITE_FILES_COMPRESSION, "zip")
            .mode(SaveMode.Append)
            .save(tempDir.toFile().getAbsolutePath());

        sparkSession.read().format(CONNECTOR_IDENTIFIER)
            .option(Options.READ_FILES_TYPE, "archive")
            .load(tempDir.toFile().getAbsolutePath())
            .show();
    }
```

src/main/java/com/marklogic/spark/reader/file/ArchiveFileReader.java
```diff
@@ -0,0 +1,75 @@
...
+    public void close() throws IOException {
+        this.zipInputStream.close();
+    }
+
+    private byte[] readZipEntry(int count) {
+        try {
+            count++;
+            byte[] content = FileUtil.readBytes(zipInputStream);
+            if(count%2 != 0 && next()){
+                return ByteArray.concat(content, FileUtil.readBytes(zipInputStream));
```
rjrudin, line 67: We don't want to concat the next entry. Instead, for an NT archive file, each time we read a zip entry, we know that the next entry will be a metadata file. So we want pseudocode like this:

```
byte[] content = readZipEntry
metadataEntry = findNextFileEntry
// Throw error if this doesn't end with ".metadata", which means it's not a valid archive file. 
byte[] metadata = readZipEntry
// Make an InternalRow using the content data and the metadata.
```

You can use `DocumentMetadataHandle.fromBuffer(byte[])` to convert the byte array into a `DocumentMetadataHandle` instance.

rjrudin (changes requested):

src/main/java/com/marklogic/spark/reader/file/ArchiveFileReader.java
rjrudin: I just started the story for reading an MLCP archive, and I think the approach I'm taking in there should be the same as what's needed here. Will share soon.

rjrudin (approved):
Looks good. We need to add support for "read.archives.categories", which I forgot to put into the story. I'll write up a separate story for that and use it as a chance to resolve the duplication across the two archive file readers.

PR #202, opened 2024-03-19
MLE-12390 Can now read MLCP archives

rjrudin:


src/main/java/com/marklogic/spark/reader/file/MlcpArchiveFileReader.java
```diff
@@ -19,39 +24,74 @@ class MlcpArchiveFileReader implements PartitionReader<InternalRow> {
...
+        this.mlcpMetadataConverter = new MlcpMetadataConverter();
     }
 
+    // TODO Add support for abortOnFailure.
     @Override
     public boolean next() throws IOException {
-        currentZipEntry = FileUtil.findNextFileEntry(zipInputStream);
-        return currentZipEntry != null;
-    }
+        // MLCP always includes a metadata entry, even if the user asks for no metadata. And the metadata entry is
```
rjrudin, line 39: @anu3990  I think you should be able to use a similar approach for reading an NT archive - though the order is opposite, as our NT archive has the content first, followed by the metadata.

Note the "TODO" as well. To support "abortOnFailure=false", we need to try building the next row in `next()`. If that fails and abortOnFailure=false, we log the error and return false. I still need to add support for that in this class.

src/test/java/com/marklogic/spark/reader/file/ReadMlcpArchiveFilesTest.java
```diff
@@ -1,43 +1,169 @@
...
+        List<Row> rows = newSparkSession().read()
+            .format(CONNECTOR_IDENTIFIER)
+            .option(Options.READ_FILES_TYPE, "mlcp_archive")
+            .load("src/test/resources/mlcp-archive-files/all-four-document-types")
+            .collectAsList();
+
+        assertEquals(4, rows.size());
+
+        assertEquals("JSON", rows.get(0).getString(2));
+        assertEquals("XML", rows.get(1).getString(2));
```
BillFarber, line 51: Looks like the test failed here.
Is collectAsList ordered?

src/main/java/com/marklogic/spark/reader/file/MlcpArchiveFileReader.java
```diff
@@ -1,61 +1,166 @@
...
+                throw new ConnectorException(message);
+            }
+            Util.MAIN_LOGGER.warn(message);
+            return new byte[0];
+        }
+    }
+
+    private InternalRow makeRow(ZipEntry contentZipEntry, byte[] content, MlcpMetadata mlcpMetadata) {
+        final String zipEntryName = contentZipEntry.getName();
+        final String uri = zipEntryName.startsWith("/") ?
```
anu3990, line 149: Can we can the private functions to a utilities class and make it accessible by other classes in the same package? These might be needed for ArchiveFileReader as well.

src/test/java/com/marklogic/spark/reader/file/ReadMlcpArchiveFilesTest.java
rjrudin: Interesting - it should be since it's a single partition reader, but maybe Spark is still free to toss those rows in the dataset in any order? I'll add a "sort" to it for the sole purpose of making the assertions reliable.

src/main/java/com/marklogic/spark/reader/file/MlcpArchiveFileReader.java
rjrudin: Definitely! But let's wait until we need it. I always like to duplicate the code first, get tests in place, and make sure it's working. Then we can do a follow-up PR that is only refactoring that gets rid of the duplication, as we have the tests to ensure we don't break anything in the process.

PR #203, opened 2024-03-19
MLE-12390 Refactoring in preparation for real PR

rjrudin:
This does 2 things:

1. Moves a bunch of logic into DocumentRowSchema so it can be reused by MlcpArchiveFileReader.
2. Enhances MlcpMetadataConverter to handle the MLCP format value. (Technically not a refactor, but it's a tiny change).

### DIFF
--- a/src/main/java/com/marklogic/spark/reader/document/DocumentRowSchema.java
+++ b/src/main/java/com/marklogic/spark/reader/document/DocumentRowSchema.java
@@ -2,12 +2,17 @@ package com.marklogic.spark.reader.document;
 
 import com.marklogic.client.io.DocumentMetadataHandle;
 import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.catalyst.util.ArrayBasedMapData;
 import org.apache.spark.sql.catalyst.util.ArrayData;
 import org.apache.spark.sql.catalyst.util.MapData;
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.StructType;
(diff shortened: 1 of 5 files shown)


src/main/java/com/marklogic/spark/reader/document/DocumentRowSchema.java
```diff
@@ -46,6 +51,61 @@ public static DocumentMetadataHandle makeDocumentMetadata(InternalRow row) {
...
+    public static void populateCollectionsColumn(Object[] row, DocumentMetadataHandle metadata) {
+        UTF8String[] collections = new UTF8String[metadata.getCollections().size()];
+        Iterator<String> iterator = metadata.getCollections().iterator();
+        for (int i = 0; i < collections.length; i++) {
+            collections[i] = UTF8String.fromString(iterator.next());
+        }
+        row[3] = ArrayData.toArrayData(collections);
+    }
+
+    public static void populatePermissionsColumn(Object[] row, DocumentMetadataHandle metadata) {
```
BillFarber, line 63: Have we talked about putting this function (or something like it) in the Java client?

BillFarber (approved):

src/main/java/com/marklogic/spark/reader/document/DocumentRowSchema.java
rjrudin: Yes - there's a `parsePermissions` method on `DocumentMetadataHandle` now, but this code is specific to Spark because of the `Object row`.

PR #204, opened 2024-03-19
MLE-12703 Can now read documents by URIs

rjrudin:


src/main/java/com/marklogic/spark/reader/document/SearchQueryBuilder.java
```diff
@@ -82,8 +84,25 @@ public SearchQueryBuilder withTransformParamsDelimiter(String delimiter) {
...
     private QueryDefinition buildQueryDefinition(DatabaseClient client) {
         final QueryManager queryManager = client.newQueryManager();
+
+        if (uris != null && uris.length > 0) {
+            StructuredQueryDefinition urisQuery = queryManager.newStructuredQueryBuilder().document(this.uris);
+            if (stringQuery != null && stringQuery.length() > 0) {
+                urisQuery.withCriteria(stringQuery);
+            }
+            if (this.query != null) {
+                Util.MAIN_LOGGER.warn("Ignoring query since a list of URIs was provided; query: {}", this.query);
```
BillFarber, line 101: Is the scenario conceivable?
A user might have a very long list of URIs and they know the want to process part of the list. Only the URIs that also match a specific criteria.
Probably far-fetched.

BillFarber (approved):

anu3990 (approved):

src/main/java/com/marklogic/spark/reader/document/SearchQueryBuilder.java
rjrudin: They can specify a subset via collections, a directory, and/or a string query. 

The challenge of supporting this, as noted in the test for it, is that we don't yet have a way of merging a StructuredQueryDefinition with the JSON and XML forms of another structured query, a serialized query, and a combined query.

PR #205, opened 2024-03-20
MLE-12390 Added support for metadata categories for archives

rjrudin:


src/test/java/com/marklogic/spark/reader/file/ReadMlcpArchiveFilesTest.java
```diff
@@ -57,6 +57,56 @@ void mlcpArchivesContainingAllFourTypesOfDocuments() {
         assertEquals("XML", rows.get(3).getString(2));
     }
 
+    @Test
+    void subsetOfCategories() {
+        readArchiveWithCategories("collections,permissions,quality").forEach(row -> {
+            verifyCollections(row);
+            verifyPermissions(row);
+            verifyQuality(row);
+            verifyColumnsAreNull(row, 6, 7);
```
BillFarber, line 66: It would be helpful to the reader if there were constants here instead of "6" or "7".

rjrudin: I was thinking about that, though they'd only be used by this test class. But yeah, I think at least adding them as constants to this test class would help.

BillFarber (approved):

PR #207, opened 2024-03-26
MLE-12958 : Read subset of metadata from NT archive

anu3990:


src/main/java/com/marklogic/spark/reader/document/DocumentRowSchema.java
```diff
@@ -114,6 +115,29 @@ public static void populateAllMetadata(Object[] row, DocumentMetadataHandle docu
         populateMetadataValuesColumn(row, documentMetadataHandle);
     }
 
+    public static void populateMetadataCategories(Object[] row, DocumentMetadataHandle documentMetadataHandle, List<String> metadataCategories) {
```
rjrudin, line 118: Putting this here creates a level of indirection that doesn't buy us anything - i.e. it's not being reused anywhere. It's better to keep it in the one place where it's used, which makes the code easier to understand. It also makes the method signature simpler since `metadataCategories` is an attribute of `ArchiveFileReader`. 

So this logic is good, but keep it in `ArchiveFileReader` until we have a need to reuse it elsewhere.

rjrudin (changes requested):

rjrudin (dismissed):

anu3990:
Rebased with the develop branch.

PR #211, opened 2024-04-01
MLE-12959 Can now use URI template on JSON content

rjrudin:
Works for "document rows" and "file rows". 

This retains the initial support for URI templates for document/file rows in the 2.2.0 release, but only when the content is not JSON. I think we can consider it an enhancement that the entire JSON document is now available for the URI template when the content is JSON.


BillFarber (approved):

BillFarber (changes requested):
Actually, there are test failures.

BillFarber (approved):

PR #215, opened 2024-04-03
MLE-12850 Can now write archive of failed documents

rjrudin:
I'm going to do a lot of manual testing with this, but it all seems straightforward - just create a `ZipFileWriter` and convert each failed `DocumentWriteOperation` into a row that the archive writer can add to a zip file as two separate entries.

### DIFF
--- a/src/main/java/com/marklogic/spark/writer/WriteBatcherDataWriter.java
+++ b/src/main/java/com/marklogic/spark/writer/WriteBatcherDataWriter.java
@@ -19,18 +19,28 @@ import com.marklogic.client.DatabaseClient;
 import com.marklogic.client.datamovement.DataMovementManager;
 import com.marklogic.client.datamovement.WriteBatcher;
 import com.marklogic.client.document.DocumentWriteOperation;
+import com.marklogic.client.impl.HandleAccessor;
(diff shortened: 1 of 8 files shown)


src/main/java/com/marklogic/spark/writer/WriteBatcherDataWriter.java
```diff
@@ -61,13 +72,22 @@ class WriteBatcherDataWriter implements DataWriter<InternalRow> {
...
         this.docBuilder = this.writeContext.newDocBuilder();
         this.databaseClient = writeContext.connectToMarkLogic();
         this.rowConverter = determineRowConverter();
-        this.batchRetrier = writeContext.isAbortOnFailure() ? null : makeBatchRetrier();
+
+        if (writeContext.isAbortOnFailure()) {
+            this.batchRetrier = null;
+            this.archiveWriter = null;
+        } else {
+            this.batchRetrier = writeContext.isAbortOnFailure() ? null : makeBatchRetrier();
```
BillFarber, line 86: Do you need this check?

BillFarber (dismissed):

BillFarber (approved):

PR #217, opened 2024-04-10
MLE-13724 Can now write arbitrary rows as XML

rjrudin:


src/main/java/com/marklogic/spark/writer/ArbitraryRowConverter.java
```diff
@@ -25,33 +27,46 @@
...
+        this.objectMapper = new ObjectMapper();
     }
 
     @Override
     public Optional<DocBuilder.DocumentInputs> convertRow(InternalRow row) {
         final String json = convertRowToJSONString(row);
-        AbstractWriteHandle contentHandle = new StringHandle(json).withFormat(Format.JSON);
-        ObjectNode columnValues = null;
+        AbstractWriteHandle contentHandle = this.xmlRootName != null ?
+            new StringHandle(convertJsonToXml(json)).withFormat(Format.JSON) :
```
BillFarber, line 51: This confuses me. If you're converting it to XML, why is Format still JSON?

rjrudin: That's a bug - though it doesn't produce a failure because the "Format" is ignored in favor of the URI extension........... making me wonder if Format should even be set. I'll change it to XML, but I may just remove it.

rjrudin: Actually, Format is worth keeping if the user chooses a URI extension that MarkLogic doesn't recognize. Would be very odd, but legit.

BillFarber (approved):

src/test/java/com/marklogic/spark/writer/WriteXmlRowsTest.java
```diff
@@ -0,0 +1,85 @@
...
+     * URI template source is a JSON representation of the row.
+     */
+    @Test
+    void uriTemplate() {
+        defaultWrite(newDefaultReader()
+            .option(Options.READ_OPTIC_QUERY, "op.fromView('Medical','Authors', '')")
+            .load()
+            .write().format(CONNECTOR_IDENTIFIER)
+            .option(Options.WRITE_COLLECTIONS, "with-uri-template")
+            .option(Options.WRITE_URI_TEMPLATE, "/xml/{ForeName}")
```
rjrudin, line 62: Made this test a little nicer by not including ".xml" in the template URI, thus ensuring that the connector is setting the Format to XML.

BillFarber (approved):

PR #220, opened 2024-04-15
MLE-13787 Ensuring "parent" filter with sqlCondition is not pushed down

rjrudin:


BillFarber (approved):

rjrudin:
Using this as a workaround until we can support converting a complex Or/Not/And filter into a single `op.sqlCondition` call.

PR #222, opened 2024-04-18
Fixed bug where success item count wasn't incremented

rjrudin:
The `BatchRetrier` was neglecting to bump up the success item count when a batch succeeded.

### DIFF
--- a/src/test/java/com/marklogic/spark/writer/WriteArchiveOfFailedDocumentsTest.java
+++ b/src/test/java/com/marklogic/spark/writer/WriteArchiveOfFailedDocumentsTest.java
@@ -6,6 +6,7 @@ import org.apache.spark.sql.Column;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SaveMode;
 import org.apache.spark.sql.SparkSession;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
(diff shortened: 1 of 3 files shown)


src/test/java/com/marklogic/spark/writer/WriteArchiveOfFailedDocumentsTest.java
```diff
@@ -46,6 +53,27 @@ void test(@TempDir Path tempDir) {
         verifyArchiveRows(rows);
     }
 
+    /**
+     * Verifies that the success count and failure counts are properly incremented.
+     */
+    @Test
+    void singlePartition(@TempDir Path tempDir) {
+        MarkLogicWrite.setSuccessCountConsumer(count -> assertEquals(1, count));
+        MarkLogicWrite.setFailureCountConsumer(count -> assertEquals(3, count));
```
BillFarber, line 62: Well, that's fascinating.
What if there's a bug in which one of those methods don't get called?

BillFarber (dismissed):
Other than my question, LGTM.

src/test/java/com/marklogic/spark/writer/WriteArchiveOfFailedDocumentsTest.java
rjrudin: Good point, I'll capture the count instead and do the assertion afterwards.

BillFarber (approved):

PR #229, opened 2024-04-30
MLE-13884 Can now read batches of triples

rjrudin:
Had to modify the `Document` classes to not assume the schema, as both the document row schema and triple row schema are now supported.


BillFarber (dismissed):

rjrudin:
Added a test for the default graph.

BillFarber (approved):

PR #233, opened 2024-05-01
MLE-13888 Can now read a naked entry in an MLCP archive

rjrudin:
Discovered that we can't yet write these though due to a bug in the Java Client. So ensuring that trying to write them doesn't cause an error for now - they're just ignored with a warning.


BillFarber (dismissed):

src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
```diff
@@ -28,8 +29,13 @@ class DocumentRowConverter implements RowConverter {
 
     @Override
     public Optional<DocBuilder.DocumentInputs> convertRow(InternalRow row) {
-        String uri = row.getString(0);
-        BytesHandle content = new BytesHandle(row.getBinary(1));
+        final String uri = row.getString(0);
+        if (row.isNullAt(2)) {
+            Util.MAIN_LOGGER.warn("Not writing document with URI {} as it has null content; this will be supported " +
```
rjrudin, line 34: Had to add this temporary fix until we do a Java Client relase.

BillFarber (approved):

PR #250, opened 2024-07-08
MLE-15472 Can now specify encoding when writing files

rjrudin:
Going to do a second PR to test encodings with archive files, figured this was enough for one PR.


src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -63,6 +63,6 @@ boolean isReadAbortOnFailure() {
 
     byte[] readBytes(InputStream inputStream) throws IOException {
         byte[] bytes = FileUtil.readBytes(inputStream);
-        return this.encoding != null ? new String(bytes).getBytes(encoding) : bytes;
+        return this.encoding != null ? new String(bytes, this.encoding).getBytes() : bytes;
```
rjrudin, line 66: This is a bugfix from the support for reading files with a specific encoding - ran into it when putting together tests in this PR. The String needs to be constructed with the encoding that the user specified.

BillFarber (approved):

PR #261, opened 2024-07-22
Can now include null fields

rjrudin:
Had to make a few changes:

1. Extracted the code for generating a Spark `JSONOptions` object along with the duplicated code for serializing a Spark row into JSON - that's all now in `JsonRowSerializer`. 
2. User can now specify `spark.marklogic.write.json.(name of option)=value` to include any option defined in Spark's `JSONOptions` class. The main one of interest is "ignoreNullFields". We'll include support for that in Flux via a new `--include-null-fields` option for some import commands (definitely for CSV). 
3. Reworked `ArbitraryRowConverter` to now happily use Jackson XML instead of the JSON library for converting JSON into an XML string.

### DIFF
--- a/src/main/java/com/marklogic/spark/writer/ArbitraryRowConverter.java
+++ b/src/main/java/com/marklogic/spark/writer/ArbitraryRowConverter.java
@@ -1,23 +1,21 @@
 package com.marklogic.spark.writer;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import com.marklogic.client.io.Format;
 import com.marklogic.client.io.JacksonHandle;
 import com.marklogic.client.io.StringHandle;
 import com.marklogic.client.io.marker.AbstractWriteHandle;
 import com.marklogic.spark.ConnectorException;
+import com.marklogic.spark.JsonRowSerializer;
 import com.marklogic.spark.Options;
-import com.marklogic.spark.Util;
 import org.apache.spark.sql.catalyst.InternalRow;
-import org.apache.spark.sql.catalyst.json.JacksonGenerator;
 import org.apache.spark.sql.types.StructField;
 import org.apache.spark.sql.types.StructType;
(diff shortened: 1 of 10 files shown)


src/main/java/com/marklogic/spark/writer/ArbitraryRowConverter.java
```diff
@@ -60,26 +57,49 @@ public Optional<DocBuilder.DocumentInputs> convertRow(InternalRow row) {
...
+        ObjectNode deserializedJson = null;
         ObjectNode uriTemplateValues = null;
-        if (this.uriTemplate != null || this.jsonRootName != null) {
-            ObjectNode jsonObject = readTree(json);
+        final boolean mustRemoveFilePathField = this.filePathIndex > 1 && jsonRowSerializer.isIncludeNullFields();
+
+        if (this.jsonRootName != null || this.xmlRootName != null || this.uriTemplate != null || mustRemoveFilePathField) {
+            deserializedJson = readTree(json);
+            if (mustRemoveFilePathField) {
+                deserializedJson.remove(MARKLOGIC_SPARK_FILE_PATH_COLUMN_NAME);
```
rjrudin, line 70: Things got a little tricky here. Once you add "ignoreNullFields=true", then the "marklogic_spark_file_path" column gets included in the JSON/XML. And we know a user doesn't want that. So this removes that column after the JSON string is deserialized into an ObjectNode. Added some tests to verify that this works correctly.

src/main/java/com/marklogic/spark/writer/ArbitraryRowConverter.java
```diff
@@ -118,33 +138,29 @@ private ObjectNode readTree(String json) {
...
      * @return
      */
-    private String convertJsonToXml(String json) {
-        JSONObject jsonObject = new JSONObject(json);
-        if (this.xmlNamespace != null) {
-            StringBuilder xml = new StringBuilder(String.format("<%s xmlns='%s'>", this.xmlRootName, this.xmlNamespace));
-            xml.append(XML.toString(jsonObject, null));
-            return xml.append(String.format("</%s>", this.xmlRootName)).toString();
+    private String convertJsonToXml(JsonNode doc) {
+        try {
```
rjrudin, line 151: Reworked to use Jackson XML. I'm not sure why I didn't do this in the first place - i.e. just replace the crappy "ObjectNode" root element with what the user wants.

BillFarber (approved):

PR #274, opened 2024-07-30
Added test for URI with non-US-ASCII characters

rjrudin:
Bumped to Java Client 7.0 to gain the fix for this.


BillFarber (approved):
The PR looks fine, but Jenkins looks like it's still having issues.

BillFarber (approved):

PR #275, opened 2024-08-01
Verifying that collatedString works

rjrudin:
Registered it as a known type so that a warning isn't logged either.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ java {
 
 repositories {
   mavenCentral()
+  mavenLocal()
   maven {
     url "https://bed-artifactory.bedford.progress.com:443/artifactory/ml-maven-snapshots/"
   }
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -21,6 +21,7 @@ java {
 
 repositories {
   mavenCentral()
+  mavenLocal()
```
BillFarber, line 24: Did you intend to check in this addition?

rjrudin: Yes, as I wasn't on the VPN and thus couldn't resolve 7.0-SNAPSHOT from Artifactory. Don't want the VPN to be a requirement for doing stuff.

src/test/java/com/marklogic/spark/reader/optic/InferSchemaTest.java
```diff
@@ -36,7 +36,7 @@ void inferFromAllTdeTypes() throws Exception {
         String columnInfoResponse = readClasspathFile("allTypes-columnInfo-response.txt");
         StructType schema = SchemaInferrer.inferSchema(columnInfoResponse);
 
-        assertEquals(35, schema.size(), "The TDE has 35 columns, and the hidden 'rowid' column that's returned " +
+        assertEquals(36, schema.size(), "The TDE has 35 columns, and the hidden 'rowid' column that's returned " +
```
BillFarber, line 39: Pretty small nit, but does the message also need to be updated?

rjrudin: Not a nit, the message is wrong. Updating.

BillFarber (approved):

PR #286, opened 2024-09-23
MLE-17041 Can now stream when reading generic files

rjrudin:
Fairly simple fix - the file is no longer read during the reader phase but rather during the writer phase to avoid reading it entirely into memory.


BillFarber (approved):

rjrudin:
We did find an issue with PUT vs POST v1/documents, will resolve that in a subsequent PR.

PR #287, opened 2024-09-23
Feature/column based params

bshalke:


rjrudin (commented):
This is a good starting point, but the immediate thing that's needed is one or more tests - not just to verify that it works, but to show how it works. An internal JIRA ticket would be useful as well to capture conversation about the problem that this solves. 

Also, is there a JIRA ticket for the server bug where you get incorrect error messages when a server timestamp is used? The only time when not using a server timestamp would be viable is when there's a single partition used and thus a single Optic call to MarkLogic. I am very reluctant to include support for anything that will almost certainly return incorrect results, even if it's for debugging. Would much rather see the server fixed.

bshalke:
I can add a test to show at least the basic functionality and an internal ticket to cover some of the use cases I've come across.

Here is the ticket with the timestamp issue - https://progresssoftware.atlassian.net/browse/MLE-14421
It looks like there hasn't been any movement on it since May.  I agree that it would be better for the issue to be fixed on the server, but right now, it's functionally impossible to diagnose any query errors without it.  I'd be okay with taking it out once the underlying server issue is fixed.

rjrudin (changes requested):
@bshalke Can you remove the workaround for getting a good error message when the timestamp is set? I definitely don't want that in the connector codebase. Also wondering if, based on Jos's feedback, the error is fixed in 11.3. 

So to move this forward:

1. Need at least one test, minimum, to show how a user would make use of this and to explain what user goal is being addressed. 
2. Remove all the code pertaining to the workaround for getting an error message when the timestamp is included.

bshalke:
Sure thing - I was at MLW this week but I will make those changes tomorrow.  The error was happening in 12.0 nightly, but I will make sure it still happens in 11.3 and RC1 and update the ticket as well.

bshalke:
Done

src/test/java/com/marklogic/spark/reader/optic/PushDownParamTest.java
```diff
@@ -0,0 +1,89 @@
...
+package com.marklogic.spark.reader.optic;
+
+import com.marklogic.spark.Options;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class PushDownParamTest extends AbstractPushDownTest {
+    static final String LEXICON_BIND_QUERY = wrapNonViewQuery(
+    "op.fromLexicons("
```
rjrudin, line 13: The connector doesn't have a hard requirement on `op.fromView`, but that's the only accessor that will scale as that's the only one we can support partitions for. 

In the Optic WG, the consensus was that parameterized CTS queries should be supported by the server, but not invented in the Spark connector. 

Is it possible to produce a test that uses `op.fromView`? I'd still consider it as an addition to the connector, but I'm generally loath to add anything that really ought to be handled by the server.

bshalke: Updated with simpler test cases that use op.fromView instead.

I agree that parameterized CTS queries should be core-server functionality.  This branch isn't about parameterizing CTS queries, but about exposing the Optic parameter bindings that already exist and are handled by the core server to downstream consumers.  Spark doesn't have a concept of parameter bindings, and neither does ODBC (which many downstream consumers use through Spark via Thrift/Hive).  Optic bindings have many use cases - term expansion, type casting, allowing cached query plans, and more.

This branch is one approach at mapping this core server functionality onto Spark's capabilities in a way that is declarative and opt-in from a Spark developer perspective, while still being agnostic and transparent from a downstream user perspective.

In addition, even if/when parameterized CTS queries are implemented by the core server, we will still need similar functionality to expose it to Spark.  This approach can be expanded in the future to support that when the capability is there.  (I will still create a PoC for it as I need it for development/exploration purposes, but that will be on a different branch and won't have a PR until/unless the core server supports it directly).

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
```diff
@@ -53,12 +54,44 @@ public class OpticReadContext extends ContextSupport {
...
+    private final Map<String, String> columnParamDefaults;
+    private Map<String, String> columnParamValues;
     private final long batchSize;
 
     public OpticReadContext(Map<String, String> properties, StructType schema, int defaultMinPartitions) {
         super(properties);
         this.schema = schema;
 
+
+        // Stash a map of the user-defined parameters (if any) so that they aren't calculated on every call.
```
rjrudin, line 67: A clean-code comment - this constructor was already a little complex. To keep the code more readable, all this new stuff should go into a new private method that initializes the different maps.

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
```diff
@@ -130,20 +172,55 @@ private PlanBuilder.Plan buildPlanForBucket(RowManager rowManager, PlanAnalysis.
             .bindParam("ML_LOWER_BOUND", bucket.lowerBound)
             .bindParam("ML_UPPER_BOUND", bucket.upperBound);
 
+        // Add user-defined parameters from options
+        for(Map.Entry<String, String> entry : this.userDefinedParams.entrySet()) {
```
rjrudin, line 176: It looks like `userDefinedParams` is not needed - yet - since the tests do not make use of this - i.e. I can remove all of this and the tests will still pass.

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
```diff
@@ -130,20 +172,55 @@ private PlanBuilder.Plan buildPlanForBucket(RowManager rowManager, PlanAnalysis.
...
+            plan = plan.bindParam(entry.getKey(), entry.getValue());
+        }
+
         if (opticFilters != null) {
             for (OpticFilter opticFilter : opticFilters) {
                 plan = opticFilter.bindFilterValue(plan);
             }
         }
 
+        for (Map.Entry<String, String> entry : columnParamValues.entrySet()) {
```
rjrudin, line 189: This should go into a new private method with a name that describes what it's doing - e.g. `bindUserParamsToPlan`. 

And I think definitely get rid of "column" in the variable names. I think `userParams` is a good term to use across the board - i.e. "op.param calls that are defined by a user in the Optic query, and which the user is responsible for providing a value for at runtime (with support for a default value as well)". The term "column" doesn't add any value.

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
```diff
@@ -130,20 +172,55 @@ private PlanBuilder.Plan buildPlanForBucket(RowManager rowManager, PlanAnalysis.
...
+            if (entry.getValue() == null) {
+                entry.setValue(columnParamDefaults.get(entry.getKey()));
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Setting default param column: {}", entry);
```
rjrudin, line 193: I think `trace` would be more appropriate for this logging. It'll get executed for every bucket and it will log the same thing every time, which is very noisy. Also, logging user-provided values can be risky - e.g. this could be a sensitive value which should not be logged. I lean towards not having this logging for now for that reason, and after we use this feature for a bit we can determine if it's really needed - perhaps we only log the param name.

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
```diff
@@ -130,20 +172,55 @@ private PlanBuilder.Plan buildPlanForBucket(RowManager rowManager, PlanAnalysis.
...
+
         return plan;
     }
 
     void pushDownFiltersIntoOpticQuery(List<OpticFilter> opticFilters) {
         this.opticFilters = opticFilters;
         // Add each filter in a separate "where" so we don't toss an op.sqlCondition into an op.and,
         // which Optic does not allow.
-        opticFilters.forEach(filter -> addOperatorToPlan(PlanUtil.buildWhere(filter)));
+        opticFilters.forEach(filter -> {
```
rjrudin, line 209: This should be another private method (lots of small private methods with self-documenting names = clean code, easy to maintain and read. Long methods = bad). 

This seems like the real key logic - i.e. check each filter in the Spark program to see if it corresponds to a user param name, and if so, add the filter value to the map that contains user params with values defined at runtime. Right?

src/test/java/com/marklogic/spark/reader/optic/PushDownParamTest.java
```diff
@@ -0,0 +1,43 @@
...
+package com.marklogic.spark.reader.optic;
+
+import com.marklogic.spark.Options;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class PushDownParamTest extends AbstractPushDownTest {
+    static final String BIND_QUERY = "op.fromView('Medical', 'Incidents', '')"
+    + ".where(fn.contains($c.Diagnosis, op.param('searchTerm')))";
```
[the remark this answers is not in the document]
bshalke: Correct - this is just intended to be a minimal example of how you could use op.param.  The param name and default value is passed via Spark option and the actual value is passed at query time via a pushed-down filter (hence the name)

src/test/java/com/marklogic/spark/reader/optic/PushDownParamTest.java
```diff
@@ -0,0 +1,43 @@
...
+     * Bindings are useful for search term expansion, transforming or casting a value,
+     * embedding in op.fromSPARQL accessors, and other advanced functionality.
+     */
+    @Test
+    void equalToWithOpticParam() {
+        assertEquals(1, newDefaultReader()
+            .option(Options.READ_OPTIC_QUERY, BIND_QUERY)
+            .option(Options.READ_COLUMN_PARAMS_PREFIX + "searchTerm", "default_value")
+            .load()
+            .filter("searchTerm == 'heat-stroke'")
```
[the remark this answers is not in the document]
bshalke: Yes, exactly.

src/test/java/com/marklogic/spark/reader/optic/PushDownParamTest.java
```diff
@@ -0,0 +1,43 @@
...
+            .collectAsList()
+            .size(), "Verifying that an optic binding can be parameterized");
```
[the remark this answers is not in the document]
bshalke: As for the tests we need - this is where I admit I'm not a Java developer, I just play one on TV.  I was just going off of the other tests you have as examples, and don't really know enough about this testing framework or what our testing standards are.

As for whether we need that new data for tests - no, I don't think so. I created it when I initially made tests for SPARQL and Lexicon queries, which is why I added it.  I also wanted to actually show off a semi-realistic use case as per our discussion, and the existing table doesn't have a collation defined on any strings so I couldn't think of a good example using them.  If we only care that the tests prove the feature works, it doesn't really matter.

src/main/java/com/marklogic/spark/Options.java
```diff
@@ -21,6 +21,8 @@ public abstract class Options {
     public static final String READ_XQUERY = "spark.marklogic.read.xquery";
     public static final String READ_XQUERY_FILE = "spark.marklogic.read.xqueryFile";
     public static final String READ_VARS_PREFIX = "spark.marklogic.read.vars.";
+    public static final String READ_PARAMS_PREFIX = "spark.marklogic.read.params.";
```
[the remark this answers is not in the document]
bshalke: This is me trying to remember what code I wrote 6 months ago, but I believe I added this first as a way to manually define a param value in options without creating a new column / have runtime pushdown support.  From my perspective, I did this so that I could build the feature out iteratively.  From a user perspective, I don't actually know how useful this is.  It could be used to test / develop queries, or maybe to reuse a single optic query for multiple downstream use cases but have different permutations that are defined with Spark options?  You can't edit Spark options on the fly, AFAIK, so I think the value is marginal at best.

rjrudin:
@bshalke  I am going to use this to fill out the details of MLE-17124 . The earliest this will go in would be Flux 1.2 (Flux 1.1 is coming out soon). Thanks for getting things moving on this front, and we'll definitely keep you involved once we use this as a starting point for implementing the feature in 1.2 or later.

src/main/java/com/marklogic/spark/Options.java
```diff
@@ -21,6 +21,8 @@ public abstract class Options {
...
+    public static final String READ_COLUMN_PARAMS_PREFIX = "spark.marklogic.read.columnParams.";
```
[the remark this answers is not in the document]
bshalke: As for the name - I agree adding optic makes sense.  I mostly just followed the pattern of what I saw already there.  The `params.{param name}={default value}` pattern is how it currently works.  I used the name columnParams to hopefully indicate that it would create a column that it uses to get the value (and to differentiate it from the hard-coded param option you commented on above), but I agree that the name isn't very communicative.  I think `spark.marklogic.read.optic.params.{name}` is fine, especially if we aren't keeping the hardcoded option.  I could also see `filterParams` or `pushdownParams` as being a better name than `columnParams`  People will have to read the documentation to figure out how to use it regardless.

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
bshalke: I will honestly have to look through my code and figure out what it's used for.  I'll let you know if I figure it out :)

src/main/java/com/marklogic/spark/reader/optic/OpticReadContext.java
bshalke: Most of the logging I added was because I was debugging something in the Java code, not because I needed it as a user, so a lot of it is probably excessive.

PR #298, opened 2024-09-26
MLE-17095 Can now stream an archive into MarkLogic

rjrudin:


src/main/java/com/marklogic/spark/reader/file/ArchiveFileReader.java
```diff
@@ -12,17 +13,20 @@
...
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.ObjectOutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
-class ArchiveFileReader implements PartitionReader<InternalRow> {
+public class ArchiveFileReader implements PartitionReader<InternalRow> {
```
BillFarber, line 24: I'm wondering if this should be two classes. One for streaming and one for not streaming. Is that even possible?

BillFarber (approved):

src/main/java/com/marklogic/spark/reader/file/ArchiveFileReader.java
rjrudin: Definitely possible. It's a choice between duplicating the workflow in another class - i.e. there's a nontrivial amount for iterating over multiple entries and multiple riles - and trying to support both streaming and non-streaming. I'll keep it in mind though - it would be a refactoring supported by the tests in place.

PR #316, opened 2024-10-09
MLE-17216 Can now split XML documents

rjrudin:
This gets the ball rolling by allowing a user to select text in an XML document and add chunks to the same document. Future stories will support JSON, different splitters, and different storage strategies. 

Going to follow this up with a small PR for Flux so that we can immediately have this available in a Flux snapshot build.

### DIFF
--- a/src/test/java/com/marklogic/spark/writer/splitter/SplitXmlDocumentTest.java
+++ b/src/test/java/com/marklogic/spark/writer/splitter/SplitXmlDocumentTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright © 2024 MarkLogic Corporation. All Rights Reserved.
+ */
+package com.marklogic.spark.writer.splitter;
+
+import com.marklogic.junit5.XmlNode;
+import com.marklogic.spark.AbstractIntegrationTest;
+import com.marklogic.spark.ConnectorException;
+import com.marklogic.spark.Options;
+import org.apache.spark.sql.DataFrameWriter;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SaveMode;
+import org.jdom2.Namespace;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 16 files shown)


src/test/java/com/marklogic/spark/writer/splitter/SplitXmlDocumentTest.java
```diff
@@ -0,0 +1,87 @@
...
+            .option(Options.CLIENT_URI, makeClientUri())
+            .option(Options.WRITE_SPLITTER_XML_PATH, "/ex:root/ex:text/text()")
+            .option(Options.WRITE_SPLITTER_XML_NAMESPACE_PREFIX + "ex", "org:example")
+            .option(Options.WRITE_PERMISSIONS, DEFAULT_PERMISSIONS)
+            .option(Options.WRITE_URI_TEMPLATE, "/namespace-test.xml")
+            .mode(SaveMode.Append)
+            .save();
+
+        XmlNode doc = readXmlDocument("/namespace-test.xml");
+        doc.setNamespaces(new Namespace[]{Namespace.getNamespace("ex", "org:example")});
```
BillFarber, line 56: Strictly out of curiosity, can you use "*" for a namespace?

rjrudin: It really depends on what the underlying jaxen library supports - https://www.cafeconleche.org/jaxen/faq.html . This will be a bit of a documentation challenge, as it doesn't support many things that MarkLogic XQuery developers are accustomed to using, such as `/fn:string()`.

BillFarber (approved):

src/test/java/com/marklogic/spark/writer/splitter/SplitXmlDocumentTest.java
rjrudin: There's also the Jakarta JAXP library - https://www.baeldung.com/java-xpath - but that shows DOM being used, which is far more difficult to use than JDOM2. But that burden is only on us. If it turns out that the XPath support is better in JAXP, we'll change to that.

PR #347, opened 2024-11-05
Verifying that DOM is a better approach for selecting chunks

rjrudin:
Just adding an initial and impl with tests showing how chunk text is selected, which aligns with what John is looking for.

Next PR will drop JDOM2 in favor of DOM.

### DIFF
--- a/src/main/java/com/marklogic/spark/writer/embedding/DOMChunk.java
+++ b/src/main/java/com/marklogic/spark/writer/embedding/DOMChunk.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright © 2024 MarkLogic Corporation. All Rights Reserved.
+ */
+package com.marklogic.spark.writer.embedding;
+
+import com.marklogic.spark.ConnectorException;
+import dev.langchain4j.data.embedding.Embedding;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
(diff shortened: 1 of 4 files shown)


src/main/java/com/marklogic/spark/writer/embedding/DOMChunk.java
```diff
@@ -0,0 +1,65 @@
...
+    public String getEmbeddingText() {
+        NodeList embeddingTextNodes;
+        try {
+            embeddingTextNodes = (NodeList) xpathFactory.newXPath().evaluate(textExpression, chunkElement, XPathConstants.NODESET);
+        } catch (XPathExpressionException e) {
+            throw new ConnectorException(String.format("Unable to evaluate XPath expression: %s; cause: %s",
+                textExpression, e.getMessage()), e);
+        }
+
+        return concatenateNotesIntoString(embeddingTextNodes);
```
BillFarber, line 47: "Notes"?

rjrudin: Fixed to Nodes.

rjrudin:
Can ignore the quality gate, as all the new code will be covered once the embedder is switched over to use the DOM approach instead of JDOM2. That'll be in the next PR.

PR #353, opened 2024-11-11
Refactored RowConverter so that an Iterator is always returned

rjrudin:
Allows for `buildAndWriteDocuments` to be used both during the connector `write` and `commit` methods. This is primarily in preparation for the next PR, which changes how the embedder batch size works.


BillFarber (dismissed):
I can't get the Jenkins job to complete successfully, so approving anyway.

PR #354, opened 2024-11-12
MLE-17715 Embedder batch size can now handle chunks across documents

rjrudin:
Previously, the batch size only worked on chunks in a document. But that's useless when a user has a small number of chunks, such as 1, per document. The batch size feature now will collect chunks from multiple documents.

This required a change in implementation. The splitter no longer knows how to add embeddings. The embedder now accepts a splitter.

So in the scenario where a user specifies both splitter and embedder options, the connector creates a splitter and then passes it to the embedder.

Also had to add support for a processor being a "supplier" of documents to be written. That allows for the embedder processor to have a pending set of chunks less than the batch size which are then processed when commit is called on the connector.


BillFarber (approved):
2 socket exceptions

BillFarber (approved):
Huzzah

PR #359, opened 2024-11-14
Refactor: Removed Spark connector references from langchain package

rjrudin:
This is to prepare for a move to multiple subprojects. The com.marklogic.langchain4j package now no longer has any references to com.marklogic.spark.

Still no functional changes, just moving furniture around.


BillFarber (approved):
Ignoring SonarQube

PR #361, opened 2024-11-14
Moved langchain4j code to new subproject

rjrudin:


marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/embedding/DocumentAndChunks.java
```diff
@@ -64,7 +63,7 @@ public String getTemporalDocumentURI() {
     }
 
     @Override
-    public int compareTo(@NotNull DocumentWriteOperation o) {
+    public int compareTo(DocumentWriteOperation o) {
```
BillFarber, line 66: Just curious as to why you removed the @NotNull.

rjrudin: It was an Intellij annotation that had been added, unbeknownst to me.

PR #364, opened 2024-11-19
Storing embedding values as floats

rjrudin:
This does not appear to have any functional impact. But given that float values are accepted, this seems preferable vs storing them as strings.


rjrudin:
Never mind, sticking with strings for now.

PR #384, opened 2024-12-16
MLE-19124 Fixed bug with failing when export path doesn't exist

rjrudin:


marklogic-spark-connector/src/main/java/com/marklogic/spark/MarkLogicFileTable.java
```diff
@@ -77,4 +84,35 @@ public Class<? extends FileFormat> fallbackFileFormat() {
...
+
+    private boolean isWriteFilesOperation(CaseInsensitiveStringMap options, Seq<String> paths) {
+        // When writing files, a user is limited to a single path. So if the user provides multiple paths when
+        // reading files, we immediately know it's not a write operation.
+        if (paths.size() != 1) {
+            return false;
+        }
+        // Unfortunately not all "read files" options have a common base. The worst case though of
+        // mis-identifying this as a "read" operation and making a directory automatically though is that
+        // the user doesn't get an expected error for trying to read a path that doesn't exist.
```
BillFarber, line 96: While this isn't terrible, it would probably be confusing.

BillFarber (approved):

marklogic-spark-connector/src/main/java/com/marklogic/spark/MarkLogicFileTable.java
rjrudin: The user will get the error logged at the "error" level if so.

PR #392, opened 2025-01-08
Getting the plumbing in place to add a Semaphore classification capability

BillFarber:


marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
```diff
@@ -15,10 +15,11 @@
...
 import com.marklogic.langchain4j.embedding.JsonChunk;
+import com.smartlogic.classificationserver.client.ClassificationException;
+import com.smartlogic.classificationserver.client.ClassificationScore;
 import dev.langchain4j.data.segment.TextSegment;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
 
 class JsonChunkDocumentProducer extends AbstractChunkDocumentProducer {
```
rjrudin, line 24: Make the two ChunkDocumentProducers receive a TextClassifier as a constructor arg instead of being part of ChunkConfig - the latter is intended to be simple options provided by a user, whereas a TextClassifier should be a constructor dependency.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
```diff
@@ -80,4 +84,24 @@ protected DocumentWriteOperation makeChunkDocument() {
     private String determineChunksArrayName(ObjectNode doc) {
         return doc.has(DEFAULT_CHUNKS_ARRAY_NAME) ? "splitter-chunks" : DEFAULT_CHUNKS_ARRAY_NAME;
     }
+
+    private ArrayNode generateConceptsForChunk(String text) {
+        // TODO - How should we handle these exceptions?
```
rjrudin, line 89: You can remove this, Sonar will complain about it otherwise, we'll revisit everything after we get feedback from S4 team.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
```diff
@@ -80,4 +84,24 @@ protected DocumentWriteOperation makeChunkDocument() {
...
+                    ObjectNode concept = conceptsArray.addObject();
+                    concept.put("id", classificationScore.getId());
+                    concept.put("name", entry.getKey());
+                    concept.put("value", classificationScore.getName());
+                    concept.put("score", classificationScore.getScore());
+                }
+            }
+            return conceptsArray;
+        } catch (ClassificationException e) {
+            throw new RuntimeException(e);
```
rjrudin, line 104: If Sonar complains about this - I think it will - change it to a ConnectorException. We'll improve error handling later.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/classifier/TextClassifier.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class TextClassifier {
+
+    private static final String THRESHOLD = "20";
+
+    private final ClassificationClient classificationClient;
+    private Token token;
+    private final TokenFetcher tokenFetcher;
+
+    public TextClassifier(String host, String protocol, String port, String endpoint, String apikey, String tokenEndpoint) {
```
rjrudin, line 23: Sonar may complain about 6 constructor args too. If so, just ignore it for now, we can add a Builder or something later.

rjrudin (approved):
All looks good, I just had comments on what I think Sonar is likely to complain about and how to fix/ignore that for now.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
rjrudin: Ignore this for now, we can refactor later.

rjrudin (dismissed):
Just one change before merging - throw the exceptions (as ConnectorException) instead of logging them. We want this to blow up right away if there's a problem. We might change that later.

PR #394, opened 2025-01-16
Change the switch for specifying the classifier protocol.

BillFarber:


rjrudin (changes requested):
Just need a change to the test to be more precise in the error it captures.

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/classifier/AddConceptsToXmlTest.java
```diff
@@ -63,6 +65,27 @@ void noConceptsAddedWhenNoSemaphoreServerSpecified() {
...
+            .option(Options.WRITE_SPLITTER_XPATH, "/root/text/text()")
+            .option(Options.WRITE_CLASSIFIER_HOST, "demo.data.progress.cloud")
+            .option(Options.WRITE_CLASSIFIER_PORT, "443")
+            .option(Options.WRITE_CLASSIFIER_ENDPOINT, "/cls/dev/cs1/")
+            .option(Options.WRITE_CLASSIFIER_APIKEY, apiKey)
+            .option(Options.WRITE_CLASSIFIER_TOKEN_ENDPOINT, "token/")
+            .mode(SaveMode.Append);
+
+        ConnectorException exception = assertThrowsConnectorException(dfw::save);
+        assertTrue(exception.getMessage().contains("Error retrieving token"));
```
rjrudin, line 86: One final nit - a problem with assertTrue/assertFalse is when they fail, like in Jenkins, it's hard to know why because those methods don't show an actual/expected like assertEquals does. So I always include an assertion message of `"Unexpected error: " + exception.getMessage()`.

rjrudin (approved):
One small thing and then merge away.

PR #395, opened 2025-01-17
Can now specify the name of the concepts array in the XML/JSON

BillFarber:


rjrudin (commented):
I'm realizing that the splitter and embedder are different in that the splitter uses "chunks" (and "splitter-chunks" if "chunks" already exists) and does not allow for customizing that. Which I like - a user can use a REST transform if needed. But the embedder does allow for the embedding element/property name to be customized. I'm not sure why I did that, as again, a REST transform can be used. 

So let's hold off on supporting this for now, as if we do it here, it will make the omission of it for chunks seem like a real gap. I'm also wondering that if we provide this, do we need to support a separate namespace for the concepts element too? I think we need a call to figure out how much flexibility is needed for all of this data generation, particularly as it differs for the splitter and embedder.

PR #396, opened 2025-01-21
Added splitter test and some messages about embedding customizations

rjrudin:
For XML, it doesn't matter if the element already exists since we're doing an `appendChild`.

For JSON - it does matter if the property name already exists. The splitter then uses "splitter-chunks" as the name. I think we can take the same approach for the classifier so we avoid the need for two more options. May later change the embedder to follow the same pattern.

### DIFF
--- a/marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/embedding/AddEmbeddingsToJsonTest.java
+++ b/marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/embedding/AddEmbeddingsToJsonTest.java
@@ -156,7 +156,10 @@ class AddEmbeddingsToJsonTest extends AbstractIntegrationTest {
         ArrayNode chunks = (ArrayNode) doc.get("envelope").get("my-chunks");
         assertEquals(2, chunks.size());
         chunks.forEach(chunk -> {
-            assertTrue(chunk.has("my-embedding"));
+            assertTrue(chunk.has("my-embedding"), "In the 2.5.0 release, both chunks and embeddings were added, " +
+                "but only embeddings can have their element name changed. It's not clear why this support was " +
+                "added instead of using the chunks approach, which is to use a name of e.g. 'embedder-embedding' " +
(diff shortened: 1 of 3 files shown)


marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/embedding/AddEmbeddingsToJsonTest.java
```diff
@@ -156,7 +156,10 @@ void customPaths() {
         ArrayNode chunks = (ArrayNode) doc.get("envelope").get("my-chunks");
         assertEquals(2, chunks.size());
         chunks.forEach(chunk -> {
-            assertTrue(chunk.has("my-embedding"));
+            assertTrue(chunk.has("my-embedding"), "In the 2.5.0 release, both chunks and embeddings were added, " +
+                "but only embeddings can have their element name changed. It's not clear why this was support was " +
```
BillFarber, line 160: "was support was"

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/embedding/AddEmbeddingsToXmlTest.java
```diff
@@ -224,7 +224,13 @@ void namespacedCustomChunksWithCustomEmbedding() {
         XmlNode doc = readXmlDocument("/split-test.xml",
             Namespace.getNamespace("ex", "org:example"), Namespace.getNamespace("ml", "org:marklogic"));
 
-        doc.assertElementCount("Each of the 2 custom chunks should have a custom embedding element.",
+        doc.assertElementCount("Each of the 2 custom chunks should have a custom embedding element. " +
+                "In the 2.5.0 release, both chunks and embeddings were added, " +
+                "but only embeddings can have their element name / namespace changed. It's not clear why this was " +
```
BillFarber, line 229: same nit

BillFarber (dismissed):

BillFarber (approved):

PR #397, opened 2025-01-22
Entire classification response saved now to both XML and JSON chunks.

BillFarber:


marklogic-langchain4j/build.gradle
```diff
@@ -11,4 +11,8 @@ dependencies {
   // Supports using Semaphore to classify documents
   implementation 'com.smartlogic.csclient:Semaphore-CS-Client:5.6.1'
   implementation 'com.smartlogic.cloud:Semaphore-Cloud-Client:5.6.1'
+  compileOnly("com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.15.2") {
```
rjrudin, line 14: This may need to be implementation - i.e. I don't think Spark typically contains this Jackson dependency.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/classifier/TextClassifier.java
```diff
@@ -29,8 +37,12 @@ public TextClassifier(String host, String https, String port, String endpoint, S
         Token token;
         try {
             token = tokenFetcher.getAccessToken();
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            builder = factory.newDocumentBuilder();
         } catch (CloudException e) {
```
rjrudin, line 42: Couple things here - a DocumentBuilder isn't thread-safe. Within the context of our connector, where a partition will have a single instance of TextClassifier, that shouldn't be a problem. But it may be a problem once we shift to a UDF. And it would definitely be a problem if we move all of this logic into a separate non-Spark-related module. So just instantiate the DBF instead. 

And related to that - `DOMHelper` already has some support in it for the annoying checked exceptions thrown by the JAXP library. Use that and rename `getDocumentBuilder` to `newDocumentBuilder` and use that instead, as it already has error handling built in.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/classifier/TextClassifier.java
```diff
@@ -29,8 +37,12 @@ public TextClassifier(String host, String https, String port, String endpoint, S
...
             throw new CloudException("Error retrieving token");
+        } catch (ParserConfigurationException e) {
```
rjrudin, line 44: Since the constructor declares that it throws `CloudException`, don't catch it, just let it propagate so that we retain the original stacktrace as well (the original error would be lost with this approach).

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/classifier/TextClassifier.java
```diff
@@ -51,17 +63,23 @@ public TextClassifier(String host, String https, String port, String endpoint, S
...
-                Util.CLASSIFIER_LOGGER.debug("Source URI: {}; count of results: {}; time: {}", sourceUri, results.size(),
-                    (System.currentTimeMillis() - start));
-            }
-            return results;
+            byte[] rawResponse = classificationClient.getClassificationServerResponse(new Body(text), new Title(sourceUri));
+            InputSource response = new InputSource(new StringReader(new String(rawResponse, StandardCharsets.UTF_8)));
+            return builder.parse(response);
         } catch (ClassificationException e) {
-            throw new MarkLogicLangchainException(String.format("Unable to generate concepts for: %s; cause: %s", text, e.getMessage()), e);
+            throw new MarkLogicLangchainException(String.format("Unable to generate classification for: %s; cause: %s", text, e.getMessage()), e);
```
rjrudin, line 72: I don't think we want to include the text in the error; the text could be of any length. Including the sourceUri should be helpful enough. 

And I don't think two catch blocks are needed. We don't want to tell a connector or Flux user that we couldn't parse the classification response as XML; there's nothing they can do about that. Just abstract it as "Unable to classify data from document with URI: sourceUri; cause: the error message". A Flux user can use `--stacktrace` to see the complete error.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
```diff
@@ -68,7 +73,15 @@ protected DocumentWriteOperation makeChunkDocument() {
             ObjectNode chunk = chunksArray.addObject();
             chunk.put("text", text);
             if (chunkConfig.getClassifier() != null) {
-                chunk.set("concepts", generateConceptsForChunk(text));
+                ObjectNode classificationObject = objectMapper.createObjectNode();
+
+                JsonNode structuredDocument = generateClassificationForChunk(text).get("STRUCTUREDDOCUMENT");
```
rjrudin, line 78: I think you can achieve the same effect with this:
```
JsonNode structuredDocument = generateClassificationForChunk(text).get("STRUCTUREDDOCUMENT");
chunk.set("classification", structuredDocument);
```

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/splitter/JsonChunkDocumentProducer.java
```diff
@@ -84,18 +97,12 @@ private String determineChunksArrayName(ObjectNode doc) {
...
-                concept.put("name", entry.getKey());
-                concept.put("value", classificationScore.getName());
-                concept.put("score", classificationScore.getScore());
-            }
+    private JsonNode generateClassificationForChunk(String text) {
+        byte[] result = chunkConfig.getClassifier().classifyTextToBytes(super.sourceDocument.getUri(), text);
+        try {
+            return xmlMapper.readTree(result);
+        } catch (IOException e) {
+            throw new MarkLogicLangchainException(String.format("Unable to generate JSON classification; cause: %s", e.getMessage()), e);
```
rjrudin, line 105: Include the sourceUri in this error message too; otherwise, a connector/Flux user won't have any idea what record(s) failed. And use the same more high-level message above - e.g. "Unable to classify data". That's all a connector/Flux user needs to know. A more advanced user can examine the stacktrace to debug it.

rjrudin (approved):
Some changes for error handling, but approving so that you can merge it if I'm already on the road this morning.

marklogic-langchain4j/build.gradle
BillFarber: I tested via Flux with a "compileOnly" build and an "implementation" build. Both worked, but I'm changing it to "implementation" anyway, just in case.

marklogic-langchain4j/src/main/java/com/marklogic/langchain4j/classifier/TextClassifier.java
BillFarber: I took a shot. I'm not sure I have what you're asking for.

BillFarber: I actually couldn't figure out what you wanted with DOMHelper, so I'm just building the DBF at instantiation and creating a new Builder during processing.

PR #401, opened 2025-01-24
Now classifies entire documents.

BillFarber:
I'm not sure if I've gone the right direction here, but reading it will give you something to do over the weekend.

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
@@ -6,10 +6,13 @@ package com.marklogic.spark.writer;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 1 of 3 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
```diff
@@ -77,12 +114,68 @@ private Iterator<DocBuilder.DocumentInputs> readContentFromRow(String uri, Inter
         if (extractedTextExists) {
             // Very primitive for now.
             final String extractedText = row.getString(8);
+
             String textUri = uri + "-extracted-text.txt";
             return Stream.of(mainInputs, new DocBuilder.DocumentInputs(textUri, new StringHandle(extractedText), uriTemplateValues, metadata)).iterator();
         }
+
+        if ((textClassifier != null) && !row.isNullAt(2)) {
+            String classificationText = new String(bytesHandle.get(), StandardCharsets.UTF_8);
```
rjrudin, line 123: You don't need to do this - both DOM and Jackson can create a Document/JsonNode from a byte array. Also, assuming UTF_8 may be an issue - there's an encoding option a user can pass in that I think you would need to use. But you can ignore that since you don't need to convert this into a string.

PR #402, opened 2025-01-26
Refactor: Added DocumentRow

rjrudin:
Intent is to avoid bloat in DocumentRowConverter and provide a nice OO wrapper around an InternalRow that represents a document. Will be useful for upcoming methods for getting chunks and classification data and embeddings from a row.


BillFarber (approved):
This refactor is really helpful.

PR #403, opened 2025-01-26
First draft classifier UDF

BillFarber:
Possibly ready for prime time.

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2025 MarkLogic Corporation. All Rights Reserved.
+ */
+package com.marklogic.spark.udf;
+
+import com.marklogic.langchain4j.classifier.TextClassifier;
+import com.smartlogic.cloud.CloudException;
+import org.apache.spark.sql.expressions.UserDefinedFunction;
+import org.apache.spark.sql.functions;
+import org.apache.spark.sql.types.DataTypes;
+
+import java.nio.charset.StandardCharsets;
+
+public abstract class TextClassifierUdf {
+
+    private static String host;
+    private static Boolean https;
+    private static String port;
+    private static String endpoint;
+    private static String apiKey;
+    private static String tokenEndpoint;
+
+    public static UserDefinedFunction build(
+        String host, Boolean https, String port, String endpoint, String apiKey, String tokenEndpoint
+    ) {
+        TextClassifierUdf.host = host;
+        TextClassifierUdf.https = https;
+        TextClassifierUdf.port = port;
+        TextClassifierUdf.endpoint = endpoint;
+        TextClassifierUdf.apiKey = apiKey;
+        TextClassifierUdf.tokenEndpoint = tokenEndpoint;
+        return functions.udf(TextClassifierUdf::classifyText, DataTypes.BinaryType);
+    }
+
+    private static byte[] classifyText(Object binaryContent) throws CloudException {
+        if (!(binaryContent instanceof byte[])) {
+            throw new IllegalArgumentException(
+                "Text classification UDF must be run against a column containing non-null byte arrays."
+            );
+        }
+        // Probably need to refactor the TextClassifier so that the token can be generated outside of that class,
+        // and passed into the constructor, so that it isn't generated on every classification call.
+        String content = new String((byte[]) binaryContent, StandardCharsets.UTF_8);
+        TextClassifier textClassifier = new TextClassifier(host, https.toString(), port, endpoint, apiKey, tokenEndpoint);
+        return textClassifier.classifyTextToBytes("sourceUri", content);
+    }
+
+    private TextClassifierUdf() {
+    }
+}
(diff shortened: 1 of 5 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
```diff
@@ -82,17 +120,97 @@ private Iterator<DocBuilder.DocumentInputs> readContentFromRow(String uri, Inter
...
+            ObjectNode originalJson = (ObjectNode) objectMapper.readTree(classificationText);
+            JsonNode structuredDocumentNode = xmlMapper.readTree(response).get(CLASSIFICATION_MAIN_ELEMENT);
+            originalJson.set("classification", structuredDocumentNode);
+            return new DocBuilder.DocumentInputs(uri, new JacksonHandle(originalJson), uriTemplateValues, metadata);
+        } catch (IOException e) {
+            throw new ConnectorException(String.format("Unable to classify data from document with URI: %s; cause: %s", uri, e.getMessage()), e);
+        }
+    }
+
+    public DocBuilder.DocumentInputs appendDocumentClassificationToXmlContent(
```
rjrudin, line 186: To keep things simple for now while we hash out the UDF approach, just support XML, ignore JSON for now. We'll need this logic for the following scenarios:

1. There's a "content" column and no "chunks" or "extractedText".
2. There's a "content" column and an "extractedText" column.
3. There's "content", "extractedText", and "chunks". 

We can do them one at a time, so this PR only needs to support the first one. But I'm thinking it'll be simpler to figure out how to support all 3 for just XML first, and then it'll be easy to add JSON support.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
```diff
@@ -82,17 +120,97 @@ private Iterator<DocBuilder.DocumentInputs> readContentFromRow(String uri, Inter
...
                 buildExtractedJsonDocument(uri, extractedText, metadata);
             return Stream.of(mainInputs, extractedTextDocument).iterator();
         }
+
+        final int classifiedTextRow = getClassifiedTextRow(row);
+        if (classifiedTextRow > 0) {
+            String classificationText = row.getString(classifiedTextRow);
+            String format = row.getString(2);
+            if ("XML".equals(format)) {
+                mainInputs = appendDocumentClassificationToXmlContent(new String(row.getBinary(1), StandardCharsets.UTF_8), uri, classificationText, uriTemplateValues, metadata);
```
rjrudin, line 137: Just paste `row.getBinary(1)` to the method, that avoids converting it into a String first.

marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
```diff
@@ -0,0 +1,51 @@
...
+            throw new IllegalArgumentException(
+                "Text classification UDF must be run against a column containing non-null byte arrays."
+            );
+        }
+        // Probably need to refactor the TextClassifier so that the token can be generated outside of that class,
+        // and passed into the constructor, so that it isn't generated on every classification call.
+        String content = new String((byte[]) binaryContent, StandardCharsets.UTF_8);
+        TextClassifier textClassifier = new TextClassifier(host, https.toString(), port, endpoint, apiKey, tokenEndpoint);
+        byte[] response = textClassifier.classifyTextToBytes("sourceUri", content);
+        return new String(response, StandardCharsets.UTF_8);
```
rjrudin, line 46: It looks simpler to just return the byte[] and toss it into the column. Granted, that's not as useful for debugging, but it avoids the String conversion since you can build a Document with DOM directly from a byte[].

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/classifier/AddClassificationToXmlTest.java
```diff
@@ -68,24 +79,47 @@ void noClassificationAddedWhenNoSemaphoreServerSpecified() {
...
+
+        assertEquals(1, dataset.count(), "Expecting 1 file");
+        assertEquals(9, dataset.collectAsList().get(0).size(),
+            "Expecting the 8 standard columns for representing a column, plus the 'textClassification' column.");
+
+        dataset.write().format(CONNECTOR_IDENTIFIER)
+            .option(Options.CLIENT_URI, makeClientUri())
+            .option(Options.WRITE_PERMISSIONS, DEFAULT_PERMISSIONS)
+            .option(Options.WRITE_URI_TEMPLATE, "/split-test.xml")
+            .option(Options.WRITE_CLASSIFIER_HOST, "demo.data.progress.cloud")
```
rjrudin, line 110: You shouldn't need any of the classifier options here since the UDF is handling it.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocumentRowConverter.java
BillFarber: > There's a "content" column and no "chunks" or "extractedText".
I'm not sure how to check for the existence of chunks.

rjrudin (commented):
Trying tossing this on top of #404, where the logic for adding the classification to the "main" document will now be in DocBuilder.

PR #407, opened 2025-01-28
Can now use a classifier UDF to classifier top-level documents with XML content.

BillFarber:
New attempt. This one uses the newly implemented DocBuilder.

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright © 2025 MarkLogic Corporation. All Rights Reserved.
+ */
+package com.marklogic.spark.udf;
+
+import com.marklogic.langchain4j.classifier.TextClassifier;
+import com.marklogic.spark.ConnectorException;
+import com.smartlogic.classificationserver.client.ClassificationConfiguration;
+import com.smartlogic.cloud.CloudException;
+import com.smartlogic.cloud.Token;
+import com.smartlogic.cloud.TokenFetcher;
+import org.apache.spark.sql.expressions.UserDefinedFunction;
+import org.apache.spark.sql.functions;
+import org.apache.spark.sql.types.DataTypes;
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class TextClassifierUdf {
+    private static final String THRESHOLD = "20";
+
+    private static ClassificationConfiguration classificationConfiguration;
+
+    public static UserDefinedFunction build(
+            String host,boolean https, String port, String endpoint, String apiKey, String tokenEndpoint
+    ) {
+        String protocol = https ? "https" : "http";
+        String tokenUrl = protocol + "://" + host + ":" + port + "/" + tokenEndpoint;
+        TokenFetcher tokenFetcher = new TokenFetcher(tokenUrl, apiKey);
+        Token token;
+        try {
+            token = tokenFetcher.getAccessToken();
+        } catch (CloudException e) {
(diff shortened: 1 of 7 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private static byte[] classifyText(Object binaryContent) {
+        if (!(binaryContent instanceof byte[])) {
+            throw new IllegalArgumentException(
+                    "Text classification UDF must be run against a column containing non-null byte arrays."
+            );
+        }
+        // Need to add handler for when a token expires midstream.
+        String content = new String((byte[]) binaryContent, StandardCharsets.UTF_8);
```
rjrudin, line 63: Can you add a Disabled test to your test class for trying this out with a document that does not have UTF-8 encoding? We have an option for encoding and it may need to be passed into here as opposed to hardcoding to UTF-8. A Disabled test is a good enough reminder.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -138,22 +171,57 @@ private DocumentWriteOperation buildMainDocument(DocumentInputs inputs) {
...
+            Document responseDoc = builder.parse(new ByteArrayInputStream(classificationResponse));
+            NodeList structuredDocumentNodeChildNodes = responseDoc.getElementsByTagName(CLASSIFICATION_MAIN_ELEMENT).item(0).getChildNodes();
+            Node classificationNode = originalDoc.createElementNS(DEFAULT_XML_NAMESPACE,"classification");
+            for (int i = 0; i < structuredDocumentNodeChildNodes.getLength(); i++) {
+                Node importedChildNode = originalDoc.importNode(structuredDocumentNodeChildNodes.item(i),true);
+                classificationNode.appendChild(importedChildNode);
+            }
+            originalDoc.getFirstChild().appendChild(classificationNode);
+
+            return new BytesHandle(convertDomToByteArray(originalDoc));
```
rjrudin, line 215: Don't need this conversion - just return a `DOMHandle(originalDoc)`. The method can return an `AbstractWriteHandle` as that's all that `DocumentWriteOperationImpl` needs. Can then remove all the Transformer stuff.

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/classifier/AddClassificationToXmlTest.java
```diff
@@ -68,24 +80,41 @@ void noClassificationAddedWhenNoSemaphoreServerSpecified() {
...
     }
 
+    @Test
+    void classifyContentsWithUdf() {
+        Dataset<Row> dataset = readDocument("/marklogic-docs/java-client-intro.xml")
+            .withColumn(CLASSIFED_TEXT_COLUMN_NAME, TEXT_CLASSIFIER.apply(new Column("content")));
+
+        assertEquals(1, dataset.count(), "Expecting 1 file");
+        assertEquals(9, dataset.collectAsList().get(0).size(),
+            "Expecting the 8 standard columns for representing a column, plus the 'textClassification' column.");
```
rjrudin, line 105: Message should refer to "classificationResponse" column.

rjrudin (changes requested):
Looks good, main change is removing the Transformer stuff as you can use a DOMHandle on a Document and can avoid the conversion to bytes.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
BillFarber: Ugh, sorry.  I remembered this as I was falling asleep last night.

PR #408, opened 2025-01-28
Can now use a classifier UDF to classifier top-level documents with JSON content.

BillFarber:


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -156,8 +160,12 @@ private DocumentWriteOperation buildMainDocument(DocumentInputs inputs) {
 
         AbstractWriteHandle content = inputs.getContent();
         byte[] classificationResponse = inputs.getClassificationResponse();
-        if (classificationResponse != null && inputs.getExtractedText() == null && Format.XML.equals(((BytesHandle) content).getFormat())) {
-            content = appendDocumentClassificationToXmlContent(((BytesHandle) content).get(), inputs.getInitialUri(), classificationResponse);
+        if (classificationResponse != null && inputs.getExtractedText() == null) {
+            if (Format.XML.equals(((BytesHandle) content).getFormat())) {
```
rjrudin: For the format check - move `determineSourceDocumentFormat` in DefaultChunkAssembler into the Util class in the same subproject. Use that here to get the format. If it returns null, log a warning, just like DefaultChunkAssembler is doing.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -156,8 +160,12 @@ private DocumentWriteOperation buildMainDocument(DocumentInputs inputs) {
...
+                content = appendDocumentClassificationToXmlContent(((BytesHandle) content).get(), inputs.getInitialUri(), classificationResponse);
+            } else if (Format.JSON.equals(((BytesHandle) content).getFormat())) {
+                content = appendDocumentClassificationToJsonContent(((BytesHandle) content).get(), inputs.getInitialUri(), classificationResponse);
```
rjrudin, line 168: The Util class in the langchain4j subproject also has a `getJsonFromHandle` method - use that here instead of casting to BytesHandle. It also handles that annoying checked exception from Jackson.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -156,8 +160,12 @@ private DocumentWriteOperation buildMainDocument(DocumentInputs inputs) {
 
         AbstractWriteHandle content = inputs.getContent();
         byte[] classificationResponse = inputs.getClassificationResponse();
-        if (classificationResponse != null && inputs.getExtractedText() == null && Format.XML.equals(((BytesHandle) content).getFormat())) {
-            content = appendDocumentClassificationToXmlContent(((BytesHandle) content).get(), inputs.getInitialUri(), classificationResponse);
+        if (classificationResponse != null && inputs.getExtractedText() == null) {
+            if (Format.XML.equals(((BytesHandle) content).getFormat())) {
+                content = appendDocumentClassificationToXmlContent(((BytesHandle) content).get(), inputs.getInitialUri(), classificationResponse);
```
rjrudin, line 166: Use DOMHelper for this. DOMHelper has an `extractDocument(DWO)` method in it - just override that with an `extractDocument(AbstractWriteHandle)` method and call it. That avoids the cast, and it also simplifies DocBuilder which won't need to depend directly on a DocumentBuilderFactory anymore.

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/classifier/AddClassificationToJsonTest.java
```diff
@@ -71,6 +88,35 @@ void noClassificationAddedWhenNoSemaphoreServerSpecified() {
         assertFalse(doc.get("chunks").get(1).has("classification"));
     }
 
+    @Test
+    @EnabledIfEnvironmentVariable(named = "SEMAPHORE_API_KEY", matches = ".*")
```
rjrudin, line 92: We want as few tests as possible to depend on Semaphore. Looking at the code, we can test everything except for actually invoking Semaphore. 

The following can be done in a follow-up PR:

1. Turn TextClassifier into an interface. 
2. Modify TextClassifierUdf to accept an instance of TextClassifier. 
3. Add tests that use a mock impl of TestClassifier, which returns a hardcoded response. 

That allows us to test the following key pieces:

1. How the response is added to the Spark row.
2. How the response is added to the 1 or more documents that are written. 

All future classifier tests should use that pattern so we can have as much coverage of the entire classifier feature without depending on an S4 connection.

rjrudin (changes requested):
Just a few changes in how the content data is retrieved - this is a common problem in other areas of the codebase, so I want to use DOMHelper and the Util class in all places. 

The note for the tests can be done in a subsequent PR.

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -155,9 +159,19 @@ private DocumentWriteOperation buildMainDocument(DocumentInputs inputs) {
         final DocumentMetadataHandle metadataFromRow = inputs.getInitialMetadata();
 
         AbstractWriteHandle content = inputs.getContent();
+        Format sourceDocumentFormat = Util.determineSourceDocumentFormat(content, sourceUri);
+        if (sourceDocumentFormat == null) {
```
rjrudin, line 163: Remove this logging, it's not relevant here. 

But it's worth having this logging below, if the classification response is not null and if the format is neither XML nor JSON - e.g. "Cannot add classification to document with URI {}; document is neither JSON nor XML." That will typically be due to a misconfiguration by the user (or by having XML/JSON documents with unrecognized extensions) and a warning is appropriate so the user isn't wondering why their documents don't have any classification data (even though it took forever to generate that classification data).

rjrudin (approved):
One logging tweak, then merge away.

PR #415, opened 2025-02-06
Can now use the Classifer UDF to classify chunks of XML & JSON

BillFarber:


marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
```diff
@@ -41,16 +44,37 @@ public static UserDefinedFunction build(
...
     }
 
-    private static byte[] classifyText(Object binaryContent) {
-        if (!(binaryContent instanceof byte[])) {
+    private static List<String> classifyText(Object binaryContent) {
+        // Need to add handler for when a token expires midstream.
+        if (binaryContent instanceof byte[]) {
+            fetchTokenIfNecessary();
+            List<String> classifications = new ArrayList<>();
+            String content = new String((byte[]) binaryContent, StandardCharsets.UTF_8);
```
rjrudin, line 55: This is fine for now but may need to become an option, as we can't assume we have UTF-8 content from a file. I can add a test later that uses a non-UTF-8 file to see what happens.

rjrudin (approved):

PR #417, opened 2025-02-06
Refactor: Removed Embedding type from Chunk interface

rjrudin:


rjrudin:
Going to reopen this, Jenkins pipeline seems to be hosed.

PR #419, opened 2025-02-06
Removed unused classifier code

rjrudin:
The factory may come back when we shift away from the UDF - but I'm expecting the factory classes to be in a different subproject at that point.


marklogic-spark-connector/src/main/java/com/marklogic/spark/udf/TextClassifierUdf.java
```diff
@@ -54,7 +54,7 @@ private static List<String> classifyText(Object binaryContent) {
             List<String> classifications = new ArrayList<>();
             String content = new String((byte[]) binaryContent, StandardCharsets.UTF_8);
             TextClassifier textClassifier = new TextClassifier(classificationConfiguration);
-            classifications.add(new String(textClassifier.classifyTextToBytes("sourceUri", content)));
+            classifications.add(new String(textClassifier.classifyTextToBytes("sourceUri", content), StandardCharsets.UTF_8));
```
BillFarber, line 57: You didn't add this to the call below on line 66. Does that mean that we can assume that chunks are already UTF-8?

BillFarber (approved):

PR #431, opened 2025-02-11
Refactor: Classifier now runs through the new processor

rjrudin:
I had to remove one test that we can't support yet - for classifying the original doc but not the chunks.

Otherwise, brought back TextClassifierFactory to replace the UDF, and moved a couple things around. DocBuilder still doing the hard work.


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/NewDocumentProcessor.java
```diff
@@ -22,25 +26,35 @@ public class NewDocumentProcessor {
...
+        if (textClassifier != null) {
             if (inputs.getExtractedText() != null) {
-                StringHandle content = new StringHandle(inputs.getExtractedText());
-                inputs.setChunks(textSplitter.split(inputs.getInitialUri(), content));
+                byte[] classification = textClassifier.classifyTextToBytes(uri, inputs.getExtractedText());
+                inputs.setClassificationResponse(classification);
             } else {
-                inputs.setChunks(textSplitter.split(inputs.getInitialUri(), inputs.getContent()));
+                String content = HandleAccessor.contentAsString(inputs.getContent());
+                byte[] classification = textClassifier.classifyTextToBytes(uri, content);
```
BillFarber, line 54: Just FYI - if we know the content is XML, we can return a Document from the classificationClient, which could be a performance improvement.

rjrudin: Looking at the S4 source code - it's just doing the same thing we are, which is feeding the byte[] into a DOM DocumentBuilder. 

I was thinking more along the lines of checking for a byte[] and avoiding the cast to a string if we have a BytesHandle. But just getting things working for now.

PR #441, opened 2025-02-13
Refactor: Removing "old" document processor approach

rjrudin:
And moving some things around. Like the new processor is now in the "core" subproject, so it'll be easier to extract in the future if/when we need to.


rjrudin:
Argh.

PR #445, opened 2025-02-14
Ensuring that the connection to S4 is closed

rjrudin:
Also removed some rogue System.out calls, and added a Gradle property for the Tika version.


rjrudin:
Going to reopen, Jenkins pipeline seems hosed.

PR #454, opened 2025-02-25
MLE-20029 Can now classify extracted text and chunks together

rjrudin:
Still need to add an option for the classifier batch size, but this use case hadn't been addressed yet.

Key changes are in `DocBuilder`, which will now add a classification to the extracted text doc if it exists.

Also swapped out one test PDF for a smaller one for easier testing.


stevebio:
One change maybe would be to ignore the SYSTEM elements in the CS response in each article unless we need them:
   <SYSTEM name="DeterminedLanguage" value="english"/>
   <SYSTEM name="LanguageGuessed" value="yes"/>
Our SOP for CS responses is to iterate the META element nodes under ARTICLE only and copying over to classification element in ML.

stevebio (approved):
Maybe in future we ignore SYSTEM elements under ARTICLE, just copy over META elements under ARTICLE which is standard approach with CS response.

PR #456, opened 2025-02-26
MLE-20029 Added option for classifier batch size

rjrudin:


stevebio (approved):
Sounds good. 20 seems reasonable, but can be modified as needed.

PR #464, opened 2025-03-14
Bug MLE-20704 - incorrect embeddings stored for chunk documents.

stevebio:
Refactor ChunkAssembler to have embeddings list passed for chunk construction.

Refactor XML and JSON producers to add the embeddings to the chunks as they are built. Add two test to ensure different text chunks do not have the same embedding vector for JSON and XML document formats.

### DIFF
--- a/marklogic-spark-api/src/main/java/com/marklogic/spark/core/splitter/AbstractChunkDocumentProducer.java
+++ b/marklogic-spark-api/src/main/java/com/marklogic/spark/core/splitter/AbstractChunkDocumentProducer.java
@@ -19,16 +19,18 @@ abstract class AbstractChunkDocumentProducer implements Iterator<DocumentWriteOp
     protected final List<String> textSegments;
     protected final ChunkConfig chunkConfig;
     protected final List<byte[]> classifications;
+    protected final List<float[]> embeddings;
     protected final int maxChunksPerDocument;
 
     protected int listIndex = -1;
     private int chunkDocumentCounter = 1;
 
-    AbstractChunkDocumentProducer(DocumentWriteOperation sourceDocument, Format sourceDocumentFormat, List<String> textSegments, ChunkConfig chunkConfig, List<byte[]> classifications) {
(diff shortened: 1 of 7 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/DocBuilder.java
```diff
@@ -310,17 +308,14 @@ private List<DocumentWriteOperation> buildChunkDocuments(DocumentInputs inputs,
             // If there's an extracted doc, we want to use that as the source document so that the user has the option
             // of adding chunks to it.
             DocumentWriteOperation sourceDocument = extractedTextDocument != null ? extractedTextDocument : mainDocument;
-            Iterator<DocumentWriteOperation> iterator = chunkAssembler.assembleChunks(sourceDocument, inputs.getChunks(), inputs.getClassifications());
+            Iterator<DocumentWriteOperation> iterator = chunkAssembler.assembleChunks(
+                sourceDocument,
+                inputs.getChunks(),
+                inputs.getClassifications(),
+                inputs.getEmbeddings());
```
rjrudin, line 315: This makes so much more sense. I'm trying to figure out why I would've done embeddings in a different manner but can't think of a reason.

marklogic-spark-api/src/main/java/com/marklogic/spark/core/splitter/AbstractChunkDocumentProducer.java
```diff
@@ -81,4 +83,15 @@ protected final String makeChunkDocumentUri(DocumentWriteOperation sourceDocumen
...
     }
+
+    /**
+     * Return the embedding at position n if it exists.
+     * @param embeddings the embeddings list
+     * @param n the position for the embedding requests
+     * @return the embedding float array
+     */
+    protected float[] getEmbeddingIfExists(List<float[]> embeddings, int n) {
+        return (embeddings != null && n < embeddings.size() ? embeddings.get(n) : null);
```
rjrudin, line 94: Remove parens around the expression, not needed.

rjrudin (approved):
Looks good! And those test failures are unrelated and stem from an internal Optic bug. Merging in, I'll remove the extra parens later.

PR #467, opened 2025-03-20
Empty file/entry no longer causes Tika to explode

rjrudin:


BillFarber (dismissed):

BillFarber (approved):
Assuming this is just a 1-off error:
[2025-03-20T20:36:17.660Z] ReadRowsTest > validPartitionCountAndBatchSize() FAILED
[2025-03-20T20:36:17.660Z]     org.opentest4j.AssertionFailedError: Expecting all 15 rows to be returned from the view ==> expected: <15> but was: <16>

PR #475, opened 2025-04-18
Building and testing against Spark 3.5.5

rjrudin:
This should be a no-op for the connector, just staying up-to-date with the latest public release of Spark. Will bump Flux next.


BillFarber (approved):

rjrudin:
Going to keep an eye on that one failure. The test passes when run by itself, and also when running the package of tests it belongs to. But it failed for me locally when running the whole suite.

PR #478, opened 2025-04-22
MLE-21194 Now classifying original doc with single article request

rjrudin:
Got some renaming to do after this - i.e. `MultiArticleClassifier` doesn't make any sense now. Will do a renaming refactor after this is merged in.


stevebio (approved):
As per our discussion, looks good.

PR #479, opened 2025-04-22
Polaris fixes

rjrudin:


marklogic-spark-connector/src/main/java/com/marklogic/spark/ReadProgressLogger.java
```diff
@@ -29,7 +30,7 @@ public static void initialize(long progressInterval, String message) {
 
     public static void logProgressIfNecessary(long itemCount) {
         if (progressInterval > 0 && progressCounter.addAndGet(itemCount) >= nextProgressInterval) {
-            synchronized (progressCounter) {
+            synchronized (lock) {
```
BillFarber, line 33: Would it make sense to check if info-level logging is enabled prior to the synchronized line. Possibly avoid the overhead and wait state if not actually logging?

BillFarber (approved):

PR #509, opened 2025-07-10
MLE-22889 Adds base64 encoding for embedding vectors

rjrudin:
Adds the option to encode embedding vectors as base64 strings, which is a more efficient way to store them in MarkLogic. This change involves:

- Updating the MarkLogic Client API dependency to a SNAPSHOT version.
- Modifying the ChunkSelector and Chunk classes to handle base64 encoding.
- Adding a new configuration option to enable/disable base64 encoding.
- Updating tests to verify the new functionality.


BillFarber (dismissed):
Who tf implemented this so cleanly?

PR #649, opened 2026-04-16
PDP-1213 Supporting a map of params for batch listener

rjrudin:
Batch listener can also optionally implement Closeable too.

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/WriteContext.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/WriteContext.java
@@ -20,6 +20,7 @@ import org.apache.spark.sql.types.StructType;
 import java.util.*;
 import java.util.function.BiConsumer;
 import java.util.function.IntConsumer;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class WriteContext extends ContextSupport {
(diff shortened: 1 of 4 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/WriteContext.java
```diff
@@ -116,25 +117,40 @@ WriteBatcher newWriteBatcher(DataMovementManager dataMovementManager) {
...
 
         return writeBatcher;
     }
 
+    private WriteBatchListener buildBatchSuccessListener() {
+        String className = getStringOption(Options.WRITE_BATCH_LISTENER_CLASSNAME);
+        try {
+            Class<?> clazz = Class.forName(className);
+            Map<String, String> params = buildBatchListenerParamsMap();
+            Object instance = clazz.getDeclaredConstructor(Map.class).newInstance(params);
```
BillFarber: What does buildBatchListenerParamsMap return if there are no properties. Will the constructor work in that case?

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/WriteWithBatchListenerTest.java
```diff
@@ -19,15 +21,26 @@ class WriteWithBatchListenerTest extends AbstractWriteTest {
 
     @Test
     void simpleBatchListener() {
-        MyBatchListener.itemCount.set(0);
+        MyBatchListener.reset();
```
BillFarber, line 24: Should there be a test for the case where the listener constructor does not take any args?

marklogic-spark-connector/src/main/java/com/marklogic/spark/writer/WriteContext.java
rjrudin: An empty map, and for now, the constructor must accept a map

marklogic-spark-connector/src/test/java/com/marklogic/spark/writer/WriteWithBatchListenerTest.java
rjrudin: Not yet since it's not supported

BillFarber (approved):

PR #650, opened 2026-04-21
PDP-1233 Added listener for commit results

rjrudin:
Needed in PDP and allows for hacky test-only consumers to be removed too.


rjrudin:
Test failures are on develop branch as well due to security changes in server.

PR #666, opened 2026-07-09
MLE-30316 Allow configurable socket timeout for classification requests in Flux

jonmille:
### Problem

The Semaphore `ClassificationConfiguration` used by the connector had a hardcoded 10-second socket timeout with no way to override it. When classifying large or complex documents (e.g., PDFs with forced language settings), server-side processing can exceed this limit, resulting in timeout errors that cannot be resolved by reducing batch size.

### Solution

Added a new `spark.marklogic.write.classifier.socketTimeout` option that lets users configure the socket timeout in milliseconds for classification requests.

When the option is not set, the Semaphore library's default of 10,000 ms (10 seconds) is preserved — no behavior change for existing users.

### Changes

**Options.java**
- Added `WRITE_CLASSIFIER_SOCKET_TIMEOUT = "spark.marklogic.write.classifier.socketTimeout"` constant (`@since 3.1.2`).

**ConfigHelper.java**
- Added `socketTimeoutMs` field, read from the connector context with a sentinel default of `-1`.
- In `buildClassificationConfiguration()`, calls `config.setSocketTimeoutMS(socketTimeoutMs)` only when the value is `> 0`, leaving the Semaphore library's default untouched otherwise.

**ConfigHelperTest.java**
- `defaultSocketTimeout()` — verifies the Semaphore library default of 10,000 ms is preserved when the option is absent.
- `customSocketTimeout()` — verifies a configured value (e.g., 30,000 ms) is applied correctly.

### Testing

Unit tests only — no MarkLogic instance required. The timeout behavior is fully verifiable by asserting against `ClassificationConfiguration.getSocketTimeoutMS()`.

```
./gradlew :marklogic-spark-connector:test --tests com.marklogic.spark.core.classifier.ConfigHelperTest
```

All 6 tests pass.

### Notes

- This option is consumed by Flux via `--classifier-timeout` (see companion PR in the `flux` repository).
- Passing `socketTimeoutMS` via `--classifier-prop` would not work as a workaround — those values are sent as HTTP query parameters to the Semaphore endpoint, not applied as client-side socket configuration.

Jira Story: https://progresssoftware.atlassian.net/browse/MLE-30316

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
@@ -31,7 +31,7 @@ class ConfigHelper {
         this.protocol = "true".equalsIgnoreCase(context.getStringOption(Options.WRITE_CLASSIFIER_HTTP)) ? "http" : "https";
         this.classifierPath = fixPath(context.getStringOption(Options.WRITE_CLASSIFIER_PATH, "/"));
         this.tokenEndpoint = fixPath(context.getStringOption(Options.WRITE_CLASSIFIER_TOKEN_PATH, "/token"));
-        this.socketTimeoutMs = context.getIntOption(Options.WRITE_CLASSIFIER_SOCKET_TIMEOUT, -1, -1);
+        this.socketTimeoutMs = context.getIntOption(Options.WRITE_CLASSIFIER_SOCKET_TIMEOUT, -1, 1);
 
         context.getProperties().forEach((key, value) -> {
             if (key.startsWith(Options.WRITE_CLASSIFIER_OPTION_PREFIX)) {
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023-2025 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+ * Copyright (c) 2023-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
  */
 package com.marklogic.spark.core.classifier;
 
@@ -22,6 +22,7 @@ class ConfigHelper {
     private final String protocol;
     private final String classifierPath;
     private final String tokenEndpoint;
+    private final int socketTimeoutMs;
     private final Map<String, String> additionalParameters = new HashMap<>();
 
     ConfigHelper(Context context) {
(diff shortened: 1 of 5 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
```diff
@@ -30,6 +31,7 @@ class ConfigHelper {
         this.protocol = "true".equalsIgnoreCase(context.getStringOption(Options.WRITE_CLASSIFIER_HTTP)) ? "http" : "https";
         this.classifierPath = fixPath(context.getStringOption(Options.WRITE_CLASSIFIER_PATH, "/"));
         this.tokenEndpoint = fixPath(context.getStringOption(Options.WRITE_CLASSIFIER_TOKEN_PATH, "/token"));
+        this.socketTimeoutMs = context.getIntOption(Options.WRITE_CLASSIFIER_SOCKET_TIMEOUT, -1, -1);
```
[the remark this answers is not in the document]
jonmille: Fixed with my own implementation.

marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
```diff
@@ -49,6 +51,9 @@ public ClassificationConfiguration buildClassificationConfiguration() {
         config.setHostPort(port);
         config.setProtocol(protocol);
         config.setHostPath(classifierPath);
+        if (socketTimeoutMs > 0) {
+            config.setSocketTimeoutMS(socketTimeoutMs);
```
rjrudin: Something to consider - while the config class receives milliseconds... how often would a user want to configure a timeout that requires specifying milliseconds? Consider a MarkLogic app server - all request timeout values are measured in seconds. 

I would expose this as a "seconds" value and then just multiply it by 1000 when calling this method. A timeout measured in seconds is easier to configure and probably what a user is expecting to provide the vast majority of the time.

rjrudin: Also, as long as we're adding this, I think it's worth adding a timeout for the connection as well.

rjrudin (changes requested):
I would do this in seconds and add an option for the connection timeout as well, just to cover both timeouts while we're doing this.

marklogic-spark-connector/src/main/java/com/marklogic/spark/core/classifier/ConfigHelper.java
jonmille: I have implemented your suggestions and added a connection timeout option as well

rjrudin (dismissed):

rjdew-progress (approved):

PR #668, opened 2026-07-10
MLE-31167 Bump jacksonVersion to 2.22.1

jonmille:
Bumping jacksonVersion from 2.21.1 to 2.22.1.

Jira Story: https://progresssoftware.atlassian.net/browse/MLE-31167


rjrudin (approved):
I highly recommend doing this in the Java Client - in fact I would bump it there first (it's on 2.22.0 and then modify this project to depend on a snapshot of the Java Client (I usually like to do a release of the Java Client before the connector/Flux for this reason).

PR #677, opened 2026-07-21
MLE-31213: Escape LIKE Wildcard Characters in Optic Filter Values to Prevent Query Injection

ngodugu-marklogic:
Escaping user-supplied values in pushed-down Optic LIKE filters now prevents wildcard and quote injection from altering query semantics.  
For StringContains, StringStartsWith, and StringEndsWith, %, _, backslash, and single quote are escaped and emitted with an ESCAPE clause, so user input is treated as literal text instead of SQL pattern/control syntax.  
In addition, string-filter attribute names are now validated against the known schema columns before interpolation, so injected column expressions are rejected immediately instead of being executed as part of a SQL condition.

### Verifying the unit tests
**1. Unit tests (no MarkLogic required)**  
Run only the new/updated unit tests in isolation:

```
cd marklogic-spark-connector  
../gradlew test --tests "com.marklogic.spark.reader.filter.FilterFactoryTest"
```

Expected result: all tests pass, including:
- stringContainsEscapesSingleQuote
- stringStartsWithEscapesSingleQuote
- stringEndsWithEscapesSingleQuote
- stringContainsEscapesLikeWildcardsAndEscapeCharacter
- stringContainsEscapesInjectionPayload
- stringContainsRejectsUnknownColumn

**2. Integration tests continue to pass (requires MarkLogic test environment)**
Run the pushdown filter suite:

`../gradlew test --tests "com.marklogic.spark.reader.optic.PushDownFilterTest"`

Expected result: existing and new filter pushdown tests pass, including:
- stringContainsNoMatch (existing baseline)
- stringContainsLiteralPercent (literal % behavior)
- stringContainsInjectionPayloadDoesNotReturnExtraRows (tautology payload no longer expands results)
- stringContainsLiteralUnderscoreMatchesOnlyLiteral
- stringContainsLiteralSingleQuoteMatchesOnlyLiteral
- stringStartsWithLiteralPercentMatchesOnlyLiteral
- stringEndsWithLiteralUnderscoreMatchesOnlyLiteral
- injectedColumnNameInStringContainsIsRejected

Notes:
- If MarkLogic is still initializing, integration tests may fail temporarily with service unavailable/retry errors; rerun once the server reports healthy.
- Non-string filter behavior is unchanged.

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/filter/FilterFactory.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/filter/FilterFactory.java
@@ -1,16 +1,22 @@
 /*
- * Copyright (c) 2023-2025 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+ * Copyright (c) 2023-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
  */
 package com.marklogic.spark.reader.filter;
 
+import com.marklogic.spark.ConnectorException;
 import org.apache.spark.sql.sources.*;
 
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public interface FilterFactory {
 
     static OpticFilter toPlanFilter(Filter filter) {
+        return toPlanFilter(filter, null);
+    }
+
+    static OpticFilter toPlanFilter(Filter filter, Set<String> knownColumnNames) {
         if (filter instanceof EqualTo f) {
             return new SingleValueFilter("eq", f.attribute(), f.value());
         } else if (filter instanceof EqualNullSafe f) {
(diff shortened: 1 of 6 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/filter/FilterFactory.java
```diff
@@ -38,12 +44,48 @@ static OpticFilter toPlanFilter(Filter filter) {
                 .map(value -> new SingleValueFilter("eq", f.attribute(), value))
                 .collect(Collectors.toList()));
         } else if (filter instanceof StringContains f) {
-            return new SqlConditionFilter(String.format("%s LIKE '%%%s%%'", f.attribute(), f.value()));
+            return new SqlConditionFilter(String.format("%s LIKE '%%%s%%' ESCAPE '\\'",
```
rjdew-progress: There are test failures because the `\` is already an escape character for both Java and JavaScript, so to get `\` to work in this scenario `\\\\` would need to be used to get the proper level of escaping. It would probably better to use a different character. Copilot  suggests using `ESCAPE '!'`

rjdew-progress (changes requested):

jonmille (approved):

rjdew-progress (approved):

PR #678, opened 2026-07-22
MLE-31216: [Spark] FIND-006: Integrate Spark Credential Redaction to Prevent Secret Exposure in Spark UI and Logs

ngodugu-marklogic:
Integrating Spark credential redaction guidance and connector-side safeguards now helps prevent MarkLogic and related secret values from being exposed in Spark UI views, event logs, and failure messages.
For spark.marklogic.client.password, spark.marklogic.write.nuclia.nuaKey, and spark.marklogic.write.classifier.apikey, the connector now documents a recommended spark.redaction.regex configuration and emits a warning when sensitive options are present but current redaction settings may not cover them.
In addition, ConnectorException now redacts sensitive values before messages are surfaced, so option-map and URI-style credential values are masked instead of being emitted in cleartext.

### Verifying the unit tests

**Unit tests (no MarkLogic required)**
Run only the new/updated unit tests in isolation:
cd marklogic-spark-connector
../gradlew test --tests "com.marklogic.spark.ConnectorExceptionTest"

Expected result: all tests pass, including:

redactsSensitiveValuesFromOptionsMapMessage
redactsPasswordInClientUriStyleMessage
redactsGenericSensitiveKeyValuePatterns
redactsMessageWhenCauseIsProvided

**Supplemental related unit test class (no MarkLogic required)**
Run connection-properties tests used during this change validation:
../gradlew test --tests "com.marklogic.spark.BuildConnectionPropertiesTest"

Expected result: all tests pass, including:

useDefaults
clientUriWithDatabase
overrideDefaults
sslEnabled
sslDisabled

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -107,6 +107,25 @@ which may be appropriate in a development or test environment. The
 [MarkLogic Java Client documentation](https://docs.marklogic.com/javadoc/client/com/marklogic/client/DatabaseClientFactory.SSLHostnameVerifier.html)
 describes the other choices for this option.
 
+## Security considerations
+
+Some connector options can contain credential or secret values, including connection-related options such as
+`spark.marklogic.client.password`.
+
+Spark may expose connector options in Spark UI pages and event logs consumed by the Spark History Server. In many
+cluster environments, the History Server is reachable by all cluster users unless additional access controls are
+configured. Without Spark redaction, credential values may appear in job metadata and error output.
+
+To reduce this risk, configure Spark redaction so that sensitive option keys are masked (such as option names
+containing `password`, `apikey`, `connectionString`, `secret`, or `token`):
+
+```
+spark.redaction.regex=(?i).*password.*|.*apikey.*|.*connectionstring.*|.*secret.*|.*token.*
+```
+
+You can apply this in Spark defaults, via `spark-submit --conf`, or in session configuration before loading data.
+This is a defense-in-depth measure; operators should also enable Spark UI authentication and TLS.
+
 ## Read options
 
 See [the guide on reading](reading-data/reading.md) for more information on how data is read from MarkLogic.
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/ContextSupport.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/ContextSupport.java
@@ -27,6 +27,7 @@ public class ContextSupport extends Context implements Serializable {
 
     public ContextSupport(Map<String, String> properties) {
         super(properties);
+        RedactionUtil.warnIfSensitiveMarkLogicOptionsMayNotBeRedacted(this);
         this.configuratorWasAdded = addOkHttpConfiguratorIfNecessary();
     }
 
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/RedactionUtil.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/RedactionUtil.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2023-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+ */
+package com.marklogic.spark;
+
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
+
+public abstract class RedactionUtil {
+
+    static final String SPARK_REDACTION_REGEX = "spark.redaction.regex";
+
+    static final String RECOMMENDED_SPARK_REDACTION_REGEX =
+        "(?i).*password.*|.*apikey.*|.*connectionstring.*|.*secret.*|.*token.*";
+
+    private static final List<String> SENSITIVE_OPTION_HINTS = Arrays.asList(
+        "password", "apikey", "secret", "token", "connectionstring"
+    );
+
+    private static final Set<String> REDACTION_WARNING_KEYS = new HashSet<>();
+
+    private RedactionUtil() {
+    }
+
+    static void warnIfSensitiveMarkLogicOptionsMayNotBeRedacted(Context context) {
+        Set<String> likelySensitiveOptions = findLikelySensitiveMarkLogicOptions(context.getProperties());
+        if (likelySensitiveOptions.isEmpty()) {
+            return;
+        }
+
+        String redactionRegex = context.getStringOption(SPARK_REDACTION_REGEX);
+        if (redactionRegex == null || redactionRegex.trim().isEmpty()) {
+            try {
+                redactionRegex = Util.getSparkSession().conf().get(SPARK_REDACTION_REGEX, null);
+            } catch (Exception ignored) {
+                // No active Spark session available, so only connector options can be inspected.
+            }
+        }
+
+        Set<String> uncoveredOptions = findUnredactedSensitiveOptions(likelySensitiveOptions, redactionRegex);
+        if (uncoveredOptions.isEmpty()) {
+            return;
+        }
+
+        final String warningKey = uncoveredOptions.stream()
+            .sorted()
+            .collect(Collectors.joining(","));
+        synchronized (REDACTION_WARNING_KEYS) {
+            if (REDACTION_WARNING_KEYS.contains(warningKey)) {
+                return;
+            }
+            REDACTION_WARNING_KEYS.add(warningKey);
+        }
+
+        Util.MAIN_LOGGER.warn(
+            "Connector options {} appear likely to contain sensitive values but may not be redacted in Spark UI/event logs. " +
+                "Set {}={} and enable Spark UI authentication/TLS for defense-in-depth.",
+            uncoveredOptions,
+            SPARK_REDACTION_REGEX,
+            RECOMMENDED_SPARK_REDACTION_REGEX
+        );
+    }
+
+    static Set<String> findLikelySensitiveMarkLogicOptions(Map<String, String> properties) {
+        return properties.entrySet().stream()
+            .filter(entry -> entry.getValue() != null && !entry.getValue().trim().isEmpty())
+            .map(Map.Entry::getKey)
+            .filter(RedactionUtil::isLikelySensitiveMarkLogicOption)
+            .collect(Collectors.toSet());
+    }
+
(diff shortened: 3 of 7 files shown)


jonmille (commented):
Looks like we have some failing tests that need to be updated to match the new credential redaction

docs/configuration.md
```diff
@@ -107,6 +107,27 @@ which may be appropriate in a development or test environment. The
 [MarkLogic Java Client documentation](https://docs.marklogic.com/javadoc/client/com/marklogic/client/DatabaseClientFactory.SSLHostnameVerifier.html)
 describes the other choices for this option.
 
+## Security considerations
+
+Some connector options contain credential values, including:
+
+- `spark.marklogic.client.password`
+- `spark.marklogic.write.nuclia.nuaKey`
```
rjrudin: These aren't publicly documented and should not be mentioned here, particularly as we ought to remove this support from the connector as we know from practice that it's not a good fit.

marklogic-spark-connector/src/main/java/com/marklogic/spark/ContextSupport.java
```diff
@@ -27,9 +31,60 @@ public class ContextSupport extends Context implements Serializable {
 
     public ContextSupport(Map<String, String> properties) {
         super(properties);
+        warnIfCredentialsMayNotBeRedacted();
         this.configuratorWasAdded = addOkHttpConfiguratorIfNecessary();
     }
 
+    private void warnIfCredentialsMayNotBeRedacted() {
```
rjrudin: I would move this logic elsewhere, it's bloating this class with some very minor logic that is rarely going to be of interest for the reader of this class.

marklogic-spark-connector/src/main/java/com/marklogic/spark/Util.java
```diff
@@ -26,6 +27,16 @@ public interface Util {
      */
     Logger MAIN_LOGGER = LoggerFactory.getLogger("com.marklogic.spark");
 
+    String SPARK_REDACTION_REGEX = "spark.redaction.regex";
+
+    List<String> CREDENTIAL_OPTION_KEYS = Arrays.asList(
```
rjrudin: This is creating a maintenance burden that a developer six months from now is unlikely to remember to update. Maybe Copilot will update it. It's also missing a number of MarkLogic connection options that are not declared in the Options class but are supported by the underlying DatabaseClientFactory newClient method that accepts a set of properties. 

That points out another maintenance burden - if a new sensitive property is added to the Java Client, it's highly unlikely that this remote section of code will be updated.

rjrudin (commented):
Is there an example of where this can actually happen? I think updating the Spark regex is a good idea to catch additional things. But I'm not convinced we need the additional code for sanitizing the messages in ConnectorException. Either Spark should have redacted the message already, or we have our own code that is improperly including potentially sensitive values in the message, in which case that code should be fixed.

rjrudin:
I did some more research on this. It sounds like if Spark builds an error message where it references options in a map, we can count on the `spark.redaction.regex` option to redact the values of certain options. 

I would not include the code though for trying to maintain a list of known sensitive options and then redacting any message passed to `ConnectorException` (CE). The list of options in this PR is not complete, and I think trying to ensure that list is complete is unlikely to be feasible. 

We should have tests covering every instance where a CE is thrown; code coverage would show where that's not the case. I think our test plumbing should be enhanced to make an additional assertion anytime a CE is caught (there are helper methods for doing this) and asserting that certain words in a potentially sensitive option name are not included.

marklogic-spark-connector/src/main/java/com/marklogic/spark/ContextSupport.java
rjrudin: After further inspection - I like the intent of this, but I would not do a static list of explicit option names. I'd instead check for any option name with "marklogic" in it along with a known set of words (e.g. `apikey`, `password, ` connectionString`) that could indicate an option name with a sensitive value. And then log any that aren't covered by the Spark redaction regex. That's a nice feature, and worst case, it's overeager and logs something that doesn't have a sensitive value.

docs/configuration.md
ngodugu-marklogic: Good call. I removed those internal option names from docs and rewrote the section to describe risk and mitigation in general terms. The guidance now focuses on Spark redaction patterns (password, apikey, connectionString, secret, token) without surfacing non-public connector options.

marklogic-spark-connector/src/main/java/com/marklogic/spark/ContextSupport.java
ngodugu-marklogic: Implemented. I replaced the explicit list with heuristic detection.

docs/configuration.md
```diff
@@ -107,6 +107,25 @@ which may be appropriate in a development or test environment. The
...
+
+Spark may expose connector options in Spark UI pages and event logs consumed by the Spark History Server. In many
+cluster environments, the History Server is reachable by all cluster users unless additional access controls are
+configured. Without Spark redaction, credential values may appear in job metadata and error output.
+
+To reduce this risk, configure Spark redaction so that sensitive option keys are masked (such as option names
+containing `password`, `apikey`, `connectionString`, `secret`, or `token`):
+
+```
+spark.redaction.regex=(?i).*password.*|.*apikey.*|.*nuakey.*|.*secret.*|.*token.*
```
rjrudin: The next step would be to set this in Flux by default.

marklogic-spark-connector/src/main/java/com/marklogic/spark/Util.java
```diff
@@ -26,6 +29,16 @@ public interface Util {
...
 
+    String SPARK_REDACTION_REGEX = "spark.redaction.regex";
+
+    List<String> SENSITIVE_OPTION_HINTS = Arrays.asList(
+        "password", "apikey", "secret", "token", "connectionstring"
+    );
+
+    Set<String> REDACTION_WARNING_KEYS = new HashSet<>();
+
+    String RECOMMENDED_SPARK_REDACTION_REGEX = "(?i).*password.*|.*apikey.*|.*nuakey.*|.*secret.*|.*token.*";
```
rjrudin: Remove nuakey from this

marklogic-spark-connector/src/main/java/com/marklogic/spark/Util.java
```diff
@@ -52,6 +65,80 @@ static boolean hasOption(Map<String, String> properties, String... options) {
             .anyMatch(option -> properties.get(option) != null && !properties.get(option).trim().isEmpty());
     }
 
+    static void warnIfSensitiveMarkLogicOptionsMayNotBeRedacted(Context context) {
```
rjrudin: This is enough code that I'd just move it to its own class now. `Util` should be a set of small, often used methods. But this is all specific to one use case. I think it's better for maintenance to dump it all into its own static class - e.g. `RedactionUtil` or something like that.

marklogic-spark-connector/src/test/java/com/marklogic/spark/AbstractIntegrationTest.java
```diff
@@ -164,7 +164,15 @@ protected final ConnectorException assertThrowsConnectorException(Runnable r) {
             "Expect the Spark-thrown SparkException to wrap our ConnectorException, which is an exception that we " +
                 "intentionally throw when an error condition is detected. " +
                 "Actual exception cause type: " + ex.getCause());
-        return (ConnectorException) ex.getCause();
+
+        ConnectorException connectorException = (ConnectorException) ex.getCause();
+        String message = connectorException.getMessage();
+        if (message != null) {
+            String lower = message.toLowerCase();
+            assertFalse(lower.matches(".*spark\\.marklogic\\.[^,\\s}\\]]*(password|apikey|secret|token|connectionstring)\\s*[=:]\\s*(?!\\*\\*\\*\\*\\*\\*\\*\\*)(\\\"[^\\\"]*\\\"|'[^']*'|[^,}\\]]+).*"),
```
rjrudin: Couldn't you instead apply the recommended Spark regex expression here and verify that it doesn't match?

marklogic-spark-connector/src/main/java/com/marklogic/spark/RedactionUtil.java
```diff
@@ -0,0 +1,100 @@
...
+ * Copyright (c) 2023-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+ */
+package com.marklogic.spark;
+
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
+
+public abstract class RedactionUtil {
```
rjrudin, line 11: One nit - assuming this is only needed by `ContextSupport`, make it non-public so it's clear to future readers that this class is not widely used.

rjrudin (approved):

jonmille (approved):

marklogic-spark-connector/src/main/java/com/marklogic/spark/RedactionUtil.java
```diff
@@ -0,0 +1,100 @@
...
+    static final String SPARK_REDACTION_REGEX = "spark.redaction.regex";
+
+    static final String RECOMMENDED_SPARK_REDACTION_REGEX =
+        "(?i).*password.*|.*apikey.*|.*connectionstring.*|.*secret.*|.*token.*";
+
+    private static final List<String> SENSITIVE_OPTION_HINTS = Arrays.asList(
+        "password", "apikey", "secret", "token", "connectionstring"
+    );
+
+    private static final Set<String> REDACTION_WARNING_KEYS = new HashSet<>();
```
RitaChen609, line 22: I suspect that the static REDACTION_WARNING_KEYS set suppresses repeated warnings across the lifetime of the JVM, which is the right production behavior — no log spam in a long-running Spark job.

It could becomes a real problem if someone later adds a test that asserts the warning was logged. In that case, the test would pass when run first in the suite but silently fail when run after another test that already populated REDACTION_WARNING_KEYS with the same key.

It's worth a comment on the PR to document this behavior so it doesn't surprise the next person who tries to write a warning-assertion test. If a future test needs to assert that the warning was logged, it will need to clear this set in @BeforeEach to avoid order-dependency with other tests.

RitaChen609 (approved):

PR #679, opened 2026-07-22
MLE-31214 added XXE hardening to DOMHelper DocumentBuilderFactory

RitaChen609:
**Summary**

Three DocumentBuilderFactory.newInstance() call sites in the connector created XML parsers with no XXE (XML External Entity) protection. Without hardening, a malicious XML document submitted to the connector could exfiltrate local files via  entity references, trigger SSRF via outbound HTTP/FTP entity resolution, or cause a DoS via recursive entity expansion (billion laughs). This is CWE-611 / OWASP A05:2021.

The fix introduces DOMHelper.newSecureDocumentBuilderFactory(), a static factory method that applies full OWASP-recommended hardening — mirroring the pattern already used by DOMHelper.newTransformerFactory() for TransformerFactory. All three unsafe call sites are updated to use it.

A Copilot code-review suggestion to also set ACCESS_EXTERNAL_DTD and ACCESS_EXTERNAL_SCHEMA (already applied to TransformerFactory) was accepted and included as defence-in-depth.

**Breaking change: XML documents containing a DOCTYPE declaration will now be rejected with a ConnectorException. This is the correct security trade-off; the user documentation has been updated accordingly.**

**Unit tests (no MarkLogic required)**

.\gradlew.bat :marklogic-spark-connector:test --tests "com.marklogic.spark.dom.DOMHelperTest" --rerun

DOMHelperTest > validXmlIsParsedSuccessfully()     PASSED
DOMHelperTest > domHelperFactoryIsNamespaceAware()  PASSED
DOMHelperTest > xmlWithDoctypeIsRejected()          PASSED
DOMHelperTest > xxeFileEntityIsRejected()           PASSED
DOMHelperTest > billionLaughsIsRejected()           PASSED
BUILD SUCCESSFUL

**XML integration suites (MarkLogic required) — no regressions**

.\gradlew.bat :marklogic-spark-connector:test \
  --tests "com.marklogic.spark.writer.splitter.SplitXmlDocumentTest" \
  --tests "com.marklogic.spark.writer.classifier.AddClassificationToXmlTest" \
  --tests "com.marklogic.spark.writer.embedding.AddEmbeddingsToXmlTest" \
  --tests "com.marklogic.spark.writer.WriteXmlRowsTest" --rerun
  
WriteXmlRowsTest             4/4   PASSED
AddClassificationToXmlTest   4/4   PASSED
AddEmbeddingsToXmlTest      14/14  PASSED
SplitXmlDocumentTest        16/16  PASSED  (1 SKIPPED — pre-existing @Disabled)
BUILD SUCCESSFUL in 58s

### DIFF
--- a/docs/writing.md
+++ b/docs/writing.md
@@ -211,6 +211,11 @@ Please see the [Flux import guide](https://marklogic.github.io/flux/import/impor
 features. While the features are primarily intended for use in Flux, they can both be used with the connector as well
 via the options described below. 
 
+**Security note:** To protect against XML External Entity (XXE) injection attacks (CWE-611), the connector's XML
+parser rejects any XML document that contains a `DOCTYPE` declaration. Attempting to write or split an XML document
+with a `DOCTYPE` declaration will result in an error. If your source XML uses a DTD, you must remove the `DOCTYPE`
+declaration before processing it with the connector.
+
 The options controlling the splitter feature are:
 
 | Option | Description | 
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/splitter/XmlChunkDocumentProducer.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/core/splitter/XmlChunkDocumentProducer.java
@@ -46,7 +46,7 @@ class XmlChunkDocumentProducer extends AbstractChunkDocumentProducer {
         this.xmlChunkConfig = new XmlChunkConfig(null, chunkConfig.getEmbeddingName(),
             chunkConfig.getEmbeddingXmlNamespace(), null, chunkConfig.isBase64EncodeVectors());
 
-        documentBuilderFactory = DocumentBuilderFactory.newInstance();
+        documentBuilderFactory = DOMHelper.newSecureDocumentBuilderFactory();
     }
 
     @Override
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/dom/DOMHelper.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/dom/DOMHelper.java
@@ -102,7 +102,11 @@ public class DOMHelper {
             factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
             factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
             factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-        } catch (ParserConfigurationException e) {
+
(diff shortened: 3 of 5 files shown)


rjrudin (commented):
Something to consider here - the Java Client has an `XmlFactories` class with public methods that produce safe factories like this. It's in an "impl" package though so it's implied to not be public.

We can of course safely depend on it because we control both projects. But I recommend doing the following:

1. Depend on the class here so we don't have to reinvent the wheel for creating a safe DOM factory.
2. Add a Jira ticket for making the class part of the public API in the Java Client. 

I think it's reasonable to add it to the public API because it's not doing anything specific to our Java Client, just generic to Java. And I imagine we have other connector/tool projects that would benefit from using those safe factory methods.

rjdew-progress (approved):

RitaChen609:
> Something to consider here - the Java Client has an `XmlFactories` class with public methods that produce safe factories like this. It's in an "impl" package though so it's implied to not be public.
> 
> We can of course safely depend on it because we control both projects. But I recommend doing the following:
> 
> 1. Depend on the class here so we don't have to reinvent the wheel for creating a safe DOM factory.
> 2. Add a Jira ticket for making the class part of the public API in the Java Client.
> 
> I think it's reasonable to add it to the public API because it's not doing anything specific to our Java Client, just generic to Java. And I imagine we have other connector/tool projects that would benefit from using those safe factory methods.

Thank you for the suggestion. We agree it's the right long-term direction and have already filed a Jira ticket to promote XmlFactories to the public API (https://progresssoftware.atlassian.net/browse/MLE-31455). However, we've decided not to introduce the dependency on com.marklogic.client.impl.XmlFactories in this PR for the duplication cannot be eliminated yet.

XmlFactories.getDocumentBuilderFactory() returns a shared, per-thread cached factory that is always namespace-aware. Two of our three call sites (XmlChunkDocumentProducer and DocBuilder) deliberately need a non-namespace-aware factory, so they cannot use the shared instance — calling setNamespaceAware(false) on it would mutate the shared state and silently break Java Client behavior on the same thread. The only path to a clean migration is makeNewDocumentBuilderFactory(), which is currently private. Until the public API ticket lands, DOMHelper.newSecureDocumentBuilderFactory() must remain anyway.

What we've done instead: DOMHelper.newSecureDocumentBuilderFactory() is now consistent with the Spark Connector's own DOMHelper.newTransformerFactory() pattern, and the filed Jira ticket captures all three call sites as candidates for migration once XmlFactories.makeNewDocumentBuilderFactory() is made public.

jonmille (approved):

ngodugu-marklogic (approved):

PR #680, opened 2026-07-22
MLE-31218: Add configurable zip bomb protection to all zip-reading paths

jonmille:
**Description:**

Addresses CWE-409 (FIND-008) from the Spark MarkLogic Connector security findings report. CVSS score: 7.5 High.

Prior to this change, no compensating controls existed for zip bomb attacks across any of the five zip-reading paths. The existing `@SuppressWarnings("java:S5042")` on `FileUtil.findNextFileEntry()` was also misplaced — that method only advances the entry iterator and never reads bytes, so the suppression had no protective effect.

Two new configurable options are added. Both default to `0` (disabled) to avoid a breaking change in this patch release — protection is opt-in:

| Option | Default | Recommended value |
|---|---|---|
| `spark.marklogic.read.zip.maxUncompressedEntryBytes` | `0` (disabled) | `268435456` (256 MB) |
| `spark.marklogic.read.zip.maxEntryCount` | `0` (disabled) | `100000` |

Any value less than `1` (including `0` and negatives) disables the limit. Set to a positive integer to enable protection. Both limits apply per archive and reset when moving to the next file in a partition.

The byte limit is enforced in `FileUtil.readBytes()` for the direct-read paths (`ZipFileReader`, `ArchiveFileReader`, `MlcpArchiveFileReader`) and via a new `MaxBytesInputStream` wrapper for the XML and RDF parser paths (`ZipAggregateXmlFileReader`, `RdfZipFileReader`), which previously consumed bytes without going through `readBytes()` at all. The entry count limit is tracked independently in each reader. `@SuppressWarnings("java:S5042")` is removed and replaced with a comment describing the two compensating controls now in place.

Streaming paths (`ZipFileIterator`, `ArchiveFileIterator`) are not covered — in streaming mode bytes flow directly to MarkLogic over the network without being buffered on the executor heap.

---

**Running the new tests**

These are integration tests requiring a running MarkLogic instance:

```bash
# Start the environment if not already running
docker compose up -d
./gradlew -i mlWaitTillReady mlDeploy

# Run only the new tests (10 total)
./gradlew :marklogic-spark-connector:test \
  --tests "com.marklogic.spark.reader.file.ReadZipBombProtectionTest"

# Run the full set of zip-related tests to confirm no regressions
./gradlew :marklogic-spark-connector:test \
  --tests "com.marklogic.spark.reader.file.ReadZipBombProtectionTest" \
  --tests "com.marklogic.spark.reader.file.ReadZipFilesTest" \
  --tests "com.marklogic.spark.reader.file.ReadAggregateXmlZipFilesTest" \
  --tests "com.marklogic.spark.reader.file.ReadRdfZipFilesTest" \
  --tests "com.marklogic.spark.reader.file.ReadMlcpArchiveFilesTest"
```

> **Note:** `ReadMlcpArchiveFilesTest.mlcpArchivesContainingAllFourTypesOfDocuments` fails on Windows due to a pre-existing Hadoop native I/O issue unrelated to this change.

---

Linked ticket: [MLE-31218](https://progresssoftware.atlassian.net/browse/MLE-31218)

[MLE-31218]: https://progresssoftware.atlassian.net/browse/MLE-31218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
@@ -135,8 +135,9 @@ public abstract class Options {
 
     /**
      * Maximum number of uncompressed bytes that may be read from a single zip archive entry.
-     * Defaults to -1 (disabled). Set to a positive integer to enable protection; a value of 268,435,456
-     * (256 MB) is a reasonable starting point for most use cases.
+     * Defaults to 0 (disabled). Any value less than 1 disables the limit. Set to a positive
+     * integer to enable protection; a value of 268,435,456 (256 MB) is a reasonable starting
+     * point for most use cases.
      *
      * @since 3.1.2
      */
@@ -144,8 +145,9 @@ public abstract class Options {
 
     /**
      * Maximum number of entries that may be iterated in a single zip archive.
-     * Defaults to -1 (disabled). Set to a positive integer to enable protection; a value of 100,000
-     * is a reasonable starting point for most use cases.
+     * Defaults to 0 (disabled). Any value less than 1 disables the limit. Set to a positive
+     * integer to enable protection; a value of 100,000 is a reasonable starting point for most
+     * use cases.
      *
      * @since 3.1.2
      */
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
@@ -135,7 +135,8 @@ public abstract class Options {
 
     /**
      * Maximum number of uncompressed bytes that may be read from a single zip archive entry.
-     * Defaults to 268,435,456 (256 MB). Set to -1 to disable the limit.
+     * Defaults to -1 (disabled). Set to a positive integer to enable protection; a value of 268,435,456
+     * (256 MB) is a reasonable starting point for most use cases.
      *
      * @since 3.1.2
      */
@@ -143,7 +144,8 @@ public abstract class Options {
 
     /**
      * Maximum number of entries that may be iterated in a single zip archive.
-     * Defaults to 100,000. Set to -1 to disable the limit.
+     * Defaults to -1 (disabled). Set to a positive integer to enable protection; a value of 100,000
+     * is a reasonable starting point for most use cases.
      *
      * @since 3.1.2
      */
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
@@ -133,6 +133,22 @@ public abstract class Options {
     public static final String READ_FILES_ABORT_ON_FAILURE = "spark.marklogic.read.files.abortOnFailure";
     public static final String READ_ARCHIVES_CATEGORIES = "spark.marklogic.read.archives.categories";
 
+    /**
+     * Maximum number of uncompressed bytes that may be read from a single zip archive entry.
+     * Defaults to 268,435,456 (256 MB). Set to -1 to disable the limit.
+     *
+     * @since 3.1.2
+     */
+    public static final String READ_ZIP_MAX_UNCOMPRESSED_ENTRY_BYTES = "spark.marklogic.read.zip.maxUncompressedEntryBytes";
+
+    /**
+     * Maximum number of entries that may be iterated in a single zip archive.
+     * Defaults to 100,000. Set to -1 to disable the limit.
+     *
+     * @since 3.1.2
+     */
+    public static final String READ_ZIP_MAX_ENTRY_COUNT = "spark.marklogic.read.zip.maxEntryCount";
+
     // "Aggregate" = an XML document containing N child elements, each of which should become a row / document.
     // "xml" is included in the name in anticipation of eventually supporting "aggregate JSON" - i.e. an array of N
     // objects, where each object should become a row / document (this is different from JSON lines format).
--- a/marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
+++ b/marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
@@ -81,38 +81,22 @@ public class FileContext extends ContextSupport implements Serializable {
 
     /**
      * @return the maximum number of uncompressed bytes to read from a single zip entry;
-     *         -1 means unlimited (the default). Zip bomb protection is opt-in: set a positive
-     *         integer to enable the limit.
-     * @throws ConnectorException if the configured value is 0 or a negative number other than -1,
-     *         since those values have no meaningful interpretation.
+     *         any value less than 1 means unlimited (the default is 0). Zip bomb protection
+     *         is opt-in: set a positive integer to enable the limit.
      */
     public long getZipMaxUncompressedEntryBytes() {
         String val = getStringOption(Options.READ_ZIP_MAX_UNCOMPRESSED_ENTRY_BYTES);
-        long limit = val != null ? Long.parseLong(val) : -1L;
-        if (limit == 0 || (limit < 0 && limit != -1L)) {
-            throw new ConnectorException(String.format(
-                "Invalid value '%s' for option '%s': must be -1 (disabled) or a positive integer.",
-                val, Options.READ_ZIP_MAX_UNCOMPRESSED_ENTRY_BYTES));
-        }
-        return limit;
+        return val != null ? Long.parseLong(val) : 0L;
     }
 
     /**
      * @return the maximum number of entries to iterate in a single zip archive;
-     *         -1 means unlimited (the default). Zip bomb protection is opt-in: set a positive
-     *         integer to enable the limit.
-     * @throws ConnectorException if the configured value is 0 or a negative number other than -1,
-     *         since those values have no meaningful interpretation.
+     *         any value less than 1 means unlimited (the default is 0). Zip bomb protection
+     *         is opt-in: set a positive integer to enable the limit.
      */
     public int getZipMaxEntryCount() {
         String val = getStringOption(Options.READ_ZIP_MAX_ENTRY_COUNT);
-        int limit = val != null ? Integer.parseInt(val) : -1;
-        if (limit == 0 || (limit < 0 && limit != -1)) {
-            throw new ConnectorException(String.format(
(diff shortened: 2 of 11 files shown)


marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -79,8 +80,52 @@ public boolean isReadAbortOnFailure() {
         return getBooleanOption(Options.READ_FILES_ABORT_ON_FAILURE, true);
     }
 
+    /**
+     * @return the maximum number of uncompressed bytes to read from a single zip entry;
+     *         -1 means unlimited. Defaults to 256 MB.
```
rjrudin: I think this is technically a breaking change - i.e. a custom may have a Flux job that is reading large trusted zip files, where an entry may be more than 256 MB. There's no problem. Then they upgrade Flux, and the job now breaks. 

For a patch or minor release, this has to default to -1l so that the feature is not enabled and thus cannot break anything that already works. 

In the next major release, it can have a default of 256 mb.

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -79,8 +80,52 @@ public boolean isReadAbortOnFailure() {
...
+        }
+        return val != null ? Long.parseLong(val) : 268_435_456L;
+    }
+
+    /**
+     * @return the maximum number of entries to iterate in a single zip archive;
+     *         -1 means unlimited. Defaults to 100,000.
+     */
+    public int getZipMaxEntryCount() {
+        String val = getStringOption(Options.READ_ZIP_MAX_ENTRY_COUNT);
```
rjrudin, line 106: Same issue as above, this would be a breaking change and thus should default to -1.

rjrudin (changes requested):

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -79,8 +80,52 @@ public boolean isReadAbortOnFailure() {
...
+    public InputStream boundedZipEntryStream(InputStream in) {
+        long maxBytes = getZipMaxUncompressedEntryBytes();
+        return maxBytes >= 0 ? new MaxBytesInputStream(in, maxBytes) : in;
+    }
+
     byte[] readBytes(InputStream inputStream) throws IOException {
-        byte[] bytes = FileUtil.readBytes(inputStream);
+        byte[] bytes = FileUtil.readBytes(inputStream, getZipMaxUncompressedEntryBytes());
         return this.encoding != null ? new String(bytes, this.encoding).getBytes() : bytes;
     }
```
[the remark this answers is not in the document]
jonmille: fixed

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/MaxBytesInputStream.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        int n = super.read(b, off, len);
+        if (n > 0) {
+            checkLimit(n);
+        }
+        return n;
+    }
+
+    /**
```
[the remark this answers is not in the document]
jonmille: fixed

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/RdfZipFileReader.java
```diff
@@ -70,6 +83,15 @@ public boolean next() {
...
+            return next();
+        } catch (ConnectorException cex) {
+            // Re-throw our own ConnectorException directly to preserve the descriptive error message.
+            // Wrapping it in a second ConnectorException would bury the actionable limit message.
+            if (fileContext.isReadAbortOnFailure()) {
+                throw cex;
+            }
+            Util.MAIN_LOGGER.warn(cex.getMessage());
             return next();
         } catch (Exception ex) {
```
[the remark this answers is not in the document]
jonmille: fixed

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/MlcpArchiveFileReader.java
```diff
@@ -126,6 +129,17 @@ private ZipEntry getNextMetadataEntry() {
             Util.MAIN_LOGGER.warn(message);
             return null;
         }
+        if (entry != null) {
+            zipEntryCount++;
+            int maxEntryCount = fileContext.getZipMaxEntryCount();
+            if (maxEntryCount >= 0 && zipEntryCount > maxEntryCount) {
+                throw new ConnectorException(String.format(
```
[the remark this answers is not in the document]
jonmille: fixed

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
jonmille: Updated the defaults to be -1

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
jonmille: Updated to default to -1

rjrudin (dismissed):
Will these become new optional args in Flux then?

jonmille:
> Will these become new optional args in Flux then?

At the time of implementation, I only measured the impact these changes would have on the flux tests. Looking back now, it would make sense to expose the same options in flux as well and default to -1 as well. I'll get something filed for implementing this in flux for 2.1.2 as well.

ngodugu-marklogic (dismissed):

rjrudin (dismissed):

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileUtil.java
```diff
@@ -25,19 +26,45 @@ public interface FileUtil {
      * @throws IOException
      */
     static byte[] readBytes(InputStream inputStream) throws IOException {
-        byte[] buffer = new byte[1024];
+        return readBytes(inputStream, -1L);
```
RitaChen609: The no-arg overload passes -1L to the new two-arg overload. But FileContext.readBytes() passes 0L for non-zip paths. Both bypass the maxBytes > 0 guard, so behavior is identical, but it introduces two sentinels for "unlimited" (0 and -1). Using 0L everywhere (matching the documented default) would be cleaner.

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -79,11 +79,61 @@ public boolean isReadAbortOnFailure() {
...
 
+    /**
+     * @return the maximum number of uncompressed bytes to read from a single zip entry;
+     *         any value less than 1 means unlimited (the default is 0). Zip bomb protection
+     *         is opt-in: set a positive integer to enable the limit.
+     */
+    public long getZipMaxUncompressedEntryBytes() {
+        String val = getStringOption(Options.READ_ZIP_MAX_UNCOMPRESSED_ENTRY_BYTES);
+        return val != null ? Long.parseLong(val) : 0L;
```
RitaChen609: No input validation. This method calls Long.parseLong on a raw option string without a try-catch.

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
```diff
@@ -79,11 +79,61 @@ public boolean isReadAbortOnFailure() {
...
+    }
+
+    /**
+     * @return the maximum number of entries to iterate in a single zip archive;
+     *         any value less than 1 means unlimited (the default is 0). Zip bomb protection
+     *         is opt-in: set a positive integer to enable the limit.
+     */
+    public int getZipMaxEntryCount() {
+        String val = getStringOption(Options.READ_ZIP_MAX_ENTRY_COUNT);
+        return val != null ? Integer.parseInt(val) : 0;
```
RitaChen609: No input validation. This method calls Integer.parseInt on a raw option string without a try-catch.

marklogic-spark-connector/src/main/java/com/marklogic/spark/Options.java
```diff
@@ -133,6 +133,26 @@ public abstract class Options {
...
 
+    /**
+     * Maximum number of uncompressed bytes that may be read from a single zip archive entry.
+     * Defaults to 0 (disabled). Any value less than 1 disables the limit. Set to a positive
+     * integer to enable protection; a value of 268,435,456 (256 MB) is a reasonable starting
+     * point for most use cases.
+     *
+     * @since 3.1.2
+     */
+    public static final String READ_ZIP_MAX_UNCOMPRESSED_ENTRY_BYTES = "spark.marklogic.read.zip.maxUncompressedEntryBytes";
```
RitaChen609, line 144: Should the new options be updated in the online docs?

jonmille: Updated the zip-files.md with zip bomb protection section

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
jonmille: Added try/catch and new test

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileContext.java
jonmille: added try/catch and new test

marklogic-spark-connector/src/main/java/com/marklogic/spark/reader/file/FileUtil.java
jonmille: Cleaned up leftover instances of -1L, replaced with 0L

RitaChen609 (approved):

